PR #43, opened 2015-07-02
Version 2 tests

abhayachauhan:
First set of functionality to build - I'll start looking at a solution when I find time next week.
Let me know if you get to it first.

### DIFF
--- a/PactNet.Tests/Mocks/MockHttpService/Comparers/ProviderServiceResponseComparerTests.cs
+++ b/PactNet.Tests/Mocks/MockHttpService/Comparers/ProviderServiceResponseComparerTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using PactNet.Matchers;
 using PactNet.Mocks.MockHttpService.Comparers;
 using PactNet.Mocks.MockHttpService.Models;
 using Xunit;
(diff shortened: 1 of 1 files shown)


abhayachauhan:
Verified the code - Looks good to start the weighing functionality

neilcampbell:
No worries.

abhayachauhan:
Hey Neil, any progress with this? I'm starting to free up some time I can revisit this.

I've heard discussions over a v3 pact spec.. we need to get aligned to v2 !

neilcampbell:
@abhayachauhan I was waiting for you on this one. I don't think we want failing tests in the mainline repo, so I didn't merge the PR. Happy to merge on a more granular level, however I think we should have tests + a specific behaviour.

steve-westwood:
How's the pattern matching coming along? Very interested in this feature!

neilcampbell:
@steve-westwood Unfortunately it's not at the moment. See https://github.com/SEEK-Jobs/pact-net/issues/9

JoannaGrzonkowska:
Hi,
You have made a really great tool!
For my master thesis I'm doing research on testing microservices and I'm extremely impressed with your work.

I'd like to ask you why https://github.com/abhayachauhan/pact-net/tree/type-matching/PactNet has not been merged into https://github.com/SEEK-Jobs/pact-net/tree/version-2 ?
I've created a nuget package from @abhayachauhan's branch and I'm successfuly testing response's body based on types. 
Is version 1.1.0 sufficient for testing your microservices?

As for the Pact Broker I must recognize that it is a fantastic idea which I have not seen in other tools. 
What I'd like to suggest is that for me it would be useful to be able to search which consumer is actually using which properties of the response body. 
For example my Provider has a method getevents, Consumer1 uses eventId and eventType, Consumer2 uses eventId, timestamp, eventType. 
I'd be interested in which consumer is actually using the timestamp property.

I've read about implementation in Rust which each language could use to perform common functions like for example validating response. Do you have any plans for Rust?

Thanks,
Asia

neilcampbell:
Hi @JoannaGrzonkowska 
Thanks for getting in touch, it's awesome to hear that you like the approach!
I didn't actually realize that the branch existed TBH, and now it's pretty stale.

Yes I have been using version 1.1.0 for testing micro services, without any huge issue. The key to doing it successfully is using provider states and having control of your test data. Version 2/3 just gives you that extra power and makes provider states less important.

That is indeed an interesting idea for the broker and definitely something that is possible. It may be worth adding that as a feature request on the Pact Broker project and if others like the idea maybe you could help out with building it? An alternative way to work that out would be to remove the timestamp property in the provider code, then run the provider tests. Any consumer that uses that property would result in a failing test.

In regards to the Rust implementation, that is where I would like to head for sure, however as with everything it's dependent on the getting the free time. It's very much an after work hours project at the moment.

Cheers,
Neil

PilchardFriendly:
Hi Neil

Just started using pact net, and 1.1.0 is good.

As others have discussed there are scenarios where we want to rely less on provider states, especially where ids and absolute URLs are in play.  Version 2 spec, in other words.

I'd be happy to pick up some work on this. I'm slightly wary of doing so if this is a part time project and you've stated that 1.1.0 is good enough for you.  Do you have a plan in mind?

neilcampbell:
@PilchardFriendly Yeah I can totally understand where you are coming from.
Don't get me wrong v2/3 support would be amazing, I just don't have the time to personally commit to it ATM. That's the beauty of open source though, anyone else can get involved and work on it.

Saying that though, I just received a v3 pull request https://github.com/SEEK-Jobs/pact-net/pull/79. Take a look at that, and see if it helps with what you are trying to achieve. Planning on reviewing it in the next day or two. If you are interested, you may be able to help out with some v2 matchers etc, based on the pact-specification https://github.com/pact-foundation/pact-specification

PR #51, opened 2015-09-18
Fix to Verify method to construct the auth header for the pact-broker url

sushant2517:
-- PactUriOptions object
-- handle the uri options in Verify method to construct auth credentials
-- unit tests


neilcampbell:
Thanks for sending this through.
Looks good! I will make a few tweaks and merge it in today sometime.

I think I will remove the support for basic auth in the uri, as it isn't needed. If we keep it, we will need to support url encoding of username/password etc.

sushant2517:
Awesome, thanks @neilcampbell. Looking forward to use the new functionality :)

PR #53, opened 2015-10-14
Schema Validation

b3hdad:
- Fixed the existing  broken unit tests
  - Added a new feature to allow verification of Schemas
  - Added unit tests for the above feature

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ Build/PactNet*.nupkg
 !packages/repositories.config
 /Build/coverage
 [Ll]ogs
+.vs/config/*
--- a/PactNet.Tests/Mocks/MockHttpService/MockProviderServiceTests.cs
+++ b/PactNet.Tests/Mocks/MockHttpService/MockProviderServiceTests.cs
@@ -134,7 +134,7 @@ namespace PactNet.Tests.Mocks.MockHttpService
         {
             var serialiserSettings = new JsonSerializerSettings
             {
-                NullValueHandling = NullValueHandling.Include,
+                NullValueHandling = NullValueHandling.Ignore,
                 StringEscapeHandling = StringEscapeHandling.Default
             };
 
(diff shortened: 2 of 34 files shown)


neilcampbell:
Hey @b3hdad
I would love to understand the intent of this change and how you are wanting to use this feature. 

A couple of additional things:
- What do you mean by "Fixed the existing broken unit tests"? 
- The AppVeyor build is failing as a result of the changes https://ci.appveyor.com/project/SEEKJobs/pact-net/build/116

b3hdad:
Basically we are trying to cover the situation where one service publishes the data in the form of json documents to a location like S3 buckets and other service(s) will consume it. So the consumers can generate a contract which specifies how the json schema should look like. Then the provider can validate a sample data against that contract. 

When I cloned the repository and ran the unit tests locally 6 of them were failing. The failures were due to serialization/deserialization. The actual was including the null values in the json while the expected was omitting the null values causing the assert to fail.

I have a look to see why the build is breaking. It passes locally when I run from visual studio. I see if I can reproduce the problem. Please bear with me as this is the first time I am dealing with Github.

neilcampbell:
Thanks mate. Let me have a think about this a little more.

None of the test should be broken to start with, as we have a CI pipeline via AppVeyor which runs all test is master to ensure everything is green. I'm not sure why they were broken on your machine.

b3hdad:
No problems. Once you review, we can discuss and see if you are happy for another change. What I am thinking is to apply similar thinking to the responses in the interactions. So one would define the pact as normal for requests and responses, however, in the response instead of specifying properties with values, we could only specify the schema and do the validation. This I think has a lot of values and makes things a lot easier.

b3hdad:
Neil,

After sharing this internally we have decided to move this out of the Pact.net and building it in its own repository/project. 

I hope you didn't spend much time on this, but please feel free to decline. 

Cheers.

neilcampbell:
@b3hdad All good mate. I was actually planning on looking into this tonight, so it's good timing!
By the sounds of what you are trying to do, version-2 of the pact specification should do something very similar. It's just a matter of someone getting the time to build it.

b3hdad:
Thanks @neilcampbell . 

I am no so sure if Pact.net version 2 will do the same thing.

Pact.net as a solution I think is for system centric architecture for dealing with synchronous operations(HTTP reqs and responses).

What I am trying to address is the data centric models where the operations are async... someone puts some data(e.g. JSON documents) somewhere(e.g. S3 bucket, DynamoDb), then whoever is interested picks it up.

I hope I make sense.

PR #55, opened 2015-10-27
Add custom reporters to pact verifier configuration

finnnewick-justgiving:
When using xunit 2 the verification log sent to the standard output isn't captured. I can work around this by setting Console.Out to something which writes to an xunit ITestOutputHelper and then putting it back to the original once the test is complete - but that doesn't play well with tests running in parallel.

So, I've added a config setting that allows adding additional reporters to be set. As the existing tests use the internal constructors that don't use the config object I've added a usage to the samples. Let me know if anything else is needed/more desirable.

### DIFF
--- a/PactNet/PactVerifierConfig.cs
+++ b/PactNet/PactVerifierConfig.cs
@@ -1,14 +1,20 @@
+using System;
+using System.Collections.Generic;
+
 namespace PactNet
 {
     public class PactVerifierConfig
     {
         public string LogDir { get; set; }
 
+        public IEnumerable<Action<string>> Reporters { get; set; }
+
         internal string LoggerName;
 
         public PactVerifierConfig()
         {
             LogDir = Constants.DefaultLogDir;
+            Reporters = new List<Action<string>>();
         }
     }
 }
--- a/PactNet/Reporters/Reporter.cs
+++ b/PactNet/Reporters/Reporter.cs
@@ -26,7 +26,7 @@ namespace PactNet.Reporters
             {
                 Console.WriteLine, 
                 new FileReportOutputter(config.LoggerName).Write
-            })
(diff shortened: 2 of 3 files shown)


neilcampbell:
@finnnewick-justgiving Thanks for the PR!

We actually added a feature a while back, which generates a log file with the output (both on the consumer and provider) for this exact reason.

Do you think this is required in addition to the log file?

The good thing about the log file, is that it pretty well just works out of the box  and the users don't need to configure anything (and we take care of thread safety). When a expectation in pact fails, we throw an exception which contains a path to the log file.

finnnewick-justgiving:
The log file is really useful but with xunit 2 we're missing an easy way to get the diff output into the test output. We could do this I guess by copying the log file into the output at the end of a test but that doesn't seem ideal.

The way I've implemented it in the pull request is to add any reporters in the config to the default ones (console and log file) rather than replace them. This means everything works as it currently does out of the box, just with the possibility of adding additional reporters.

neilcampbell:
Ok cool, just wanted to make sure you saw the log file, otherwise we would have to make that more explicit.

Let me have a bit more of a think about this one, and see if we can formalise this concept a little further.

neilcampbell:
Really sorry this has taken so long to be merged, I have been flat out the last few months.

I have added the concept of IReportOutputter, so you can now create your own report outputter. See below.

```c#
var outputter = new CustomOutputter(); //Implements IReportOutputter
var config = new PactVerifierConfig();
config.ReportOutputters.Add(outputter); //Add your custom report outputter
config.ReportOutputters.RemoveAt(0); //Optionally you can remove a default report outputter if you want
```

I will let you know once a nuget package has been produced.

neilcampbell:
A new nuget package has been published, 1.1.0 contains this feature.

PR #60, opened 2016-04-20
Value agnostic http body matcher

ceddlyburge:
HI there

I have added the ability for the client to specify that it does not care about the property values (JValues) in the response body, but that it still cares that the existence of the properties. This makes our tests simpler to setup and more robust.

Being as this is specified by the client, I had to modify the json PactFile format, which I realise might be controversial, but the new value is optional and defaults to the existing behaviour so hopefully it won't be a big deal.

The new property might ideally go on the body / ProviderServiceResponse.Body properties, but this would result in a reserved property name, so it has gone on the Interaction object instead. This also removed the mild hack of creating MatchingRules while deserialising ProviderServiceResponse.Body. I have added an overload of Compare to IProviderServiceResponseComparer, which is mildly suboptimal and means that you would lose the value agnostic functionality if treating it as an IComparer. However nothing does this and there doesn't seem to be any reason why anything would, so I don't think its a problem.

I have packaged this change up as v1.1.2 and added it to our internal NuGet feed, so if you decide not to accept this pull request, could you do us a favour and jump to version 1.1.3 when you next release?

Thanks

Cedd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ Build/PactNet*.nupkg
 !packages/repositories.config
 /Build/coverage
 [Ll]ogs
+.vs/
--- a/PactNet.Tests/Mocks/MockHttpService/Comparers/ProviderServiceResponseComparerTests.cs
+++ b/PactNet.Tests/Mocks/MockHttpService/Comparers/ProviderServiceResponseComparerTests.cs
@@ -379,6 +379,39 @@ namespace PactNet.Tests.Mocks.MockHttpService.Comparers
         }
 
         [Fact]
+        public void Compare_ValueAgnostic_WithNonMatchingObject_OneErrorIsAddedToTheComparisonResult()
+        {
+            var expected = new ProviderServiceResponse
+            {
+                Status = 201,
+                Body = new
+                {
+                    myString = "Tester",
+                    myInt = 1,
+                    myGuid = Guid.Parse("EEB517E6-AC8B-414A-A0DB-6147EAD9193C"),
+                    myDouble = 2.0
+                }
+            };
+
+            var actual = new ProviderServiceResponse
+            {
+                Status = 201,
+                Body = new
+                {
+                    myString = "Tester",
+                    MyInt = 1,
+                    MyGuid = Guid.Parse("EEB517E6-AC8B-414A-A0DB-6147EAD9193C")
+                }
+            };
+
+            var comparer = GetSubject();
+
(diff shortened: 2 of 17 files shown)


neilcampbell:
I'm not sure if you have seen the discussion about the Pact v2 specification and flexible value matching? Basically we cannot implement any custom matching that does not conform to the Pact specification (https://github.com/pact-foundation/pact-specification/tree/version-2). The reason being that there are multiple language implementations that should be cross compatible. So for example a .NET consumer talking to a Ruby provider, or a JavaScript consumer talking to a .NET provider. Currently the .NET version does not support v2, and I just don't have the time to implement this (supporting this project in what little spare time I have). If you are interested in reading more see https://github.com/SEEK-Jobs/pact-net/issues/9. If you are interested in helping out, that would be super awesome and I would be really grateful! I have done some of the work (on branch version-2). A few others have mentioned they want to help, but I am yet to hear anything back.

ceddlyburge:
Hi Neil

Is that v2 specification finalised?

The matching rules in the example below seems closest to what I want:

"matchingRules": { "$.body.alligator.name": {"match": "type"}, "
$.body.alligator.feet": {"match": "type"} }
Although I would want to be able to specify this for all properties,
instead of specifying each property individually.

So if I would want to make it look like this

"matchingRules": { "match": "type" }

or this

"matchingRules": { "all": {"match": "type"} }

or this (very similar to what I now have)

"matchingRules": { "match": "propertyName" }

At one point in #9 <https://github.com/SEEK-Jobs/pact-net/issues/9> it says
that this is possible by applying the match on the root property and
letting it filter down, although there is no test case that shows it that I
could see, and maybe it never made the final cut.

Anyway, that v2 specification looks to be very different from v1, and I
don't have time for all of that. I have basically done some flexible
matching already though, so if it gets to a point where I can some small
bit in isolation than I may be able to help.

It seems that this is going to be quite a long way off, so we will probably
carry on with what we have now as a pragmatic solution.

Cheers

Cedd

neilcampbell:
Yeah you should be able to apply the match at the root and that should filter down.
Fair enough about using your own branch (that the beauty of open source), just know that it will not be compatible or supported with new versions of PactNet or the other language variants.

As such I cannot merge this PR unfortunately.

ceddlyburge:
Hi Neil, no worries.

I might call this version 1.2 so that it doesn't clash with anything that
already exists. Good luck with implementing v2, it looks like a big chunk
of work.

Cheers

Cedd

PR #62, opened 2016-05-02
Fixing the dropping of Content-Type header parameters when verifying …

gruntster:
The Verifier drops parameters from the Content-Type header, e.g. the header `application/json; version=1; charset=utf-8` would currently be submitted as `application/json; charset=utf-8`.  (Pacts are generated correctly by the Builder though when the Content-Type header contains parameters)


neilcampbell:
Thanks @gruntster! Will have a look at this tonight and hopefully get it merged in for you.

neilcampbell:
@gruntster I have merged the PR into master and the fix is available in version 1.1.2 nuget package

PR #63, opened 2016-05-05
Verifier tweaks

gruntster:
- handle content type parameters separately from the media type to avoid misinterpretations
- support all charset encodings available in .NET rather than just Unicode variants and ASCII
- Verifier preserves original casing of the Content-Type
- use latest 1.1 specification tests


neilcampbell:
Awesome stuff! Will have a look at this over the weekend.

neilcampbell:
PR has been merged and is available in Nuget Package 1.1.3. Sorry this took a bit longer than expected, it's been a crazy week!

PR #66, opened 2016-07-09
Adding related tools section to the readme files.

b3hdad:
Since the two pact based tools/libs have been open sourced, and they are also .net based, I thought I might be a good idea to mentioned them here.


neilcampbell:
Hey @b3hdad!
Nice one, merged in.

b3hdad:
Hi @neilcampbell,
that's awesome thank you.

PR #70, opened 2016-11-01
Add rewriteLocalhost overload for disabling netsh localhost binding s…

vdex42:
…o that pact.net can be run without admin privileges

By default nancy runs something like **netsh http add urlacl url=http://+:8799** in order to bind to all ip addresses and not just localhost. The problem with this is that it requires admin privileges to run and makes automation on build servers etc a lot harder.

Nancy has a RewriteLocalhost setting that can disable this, which means only calls to "localhost" will work, but won't require admin privileges to run. 
https://github.com/NancyFx/Nancy/wiki/Self-Hosting-Nancy#host-configuration

### DIFF
--- a/PactNet/IPactBuilder.cs
+++ b/PactNet/IPactBuilder.cs
@@ -8,8 +8,7 @@ namespace PactNet
         IPactBuilder ServiceConsumer(string consumerName);
         IPactBuilder HasPactWith(string providerName);
         IMockProviderService MockService(int port, bool enableSsl = false);
-        IMockProviderService MockService(int port, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false);
-        IMockProviderService MockService(int port, bool rewriteLocalhost, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false);
+        IMockProviderService MockService(int port, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false, bool bindOnAllAdapters = false);        
         void Build();
     }
 }
--- a/PactNet/IPactBuilder.cs
+++ b/PactNet/IPactBuilder.cs
@@ -9,6 +9,7 @@ namespace PactNet
         IPactBuilder HasPactWith(string providerName);
         IMockProviderService MockService(int port, bool enableSsl = false);
         IMockProviderService MockService(int port, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false);
+        IMockProviderService MockService(int port, bool rewriteLocalhost, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false);
         void Build();
     }
 }
--- a/PactNet/Mocks/MockHttpService/MockProviderService.cs
+++ b/PactNet/Mocks/MockHttpService/MockProviderService.cs
@@ -41,7 +41,7 @@ namespace PactNet.Mocks.MockHttpService
             _httpMethodMapper = httpMethodMapper;
         }
 
-        public MockProviderService(int port, bool enableSsl, bool bindOnAllAdapters, string providerName, PactConfig config)
+        public MockProviderService(int port, bool enableSsl, string providerName, PactConfig config, bool bindOnAllAdapters)
             : this(
             baseUri =>
             {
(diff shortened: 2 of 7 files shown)


neilcampbell:
@vdex42 Nice, thanks for this!
I actually think this should be the default option, as it is the most common scenario.

I have added a few code review comments inline for you.

PactNet/IPactBuilder.cs
```diff
@@ -9,6 +9,7 @@ public interface IPactBuilder
         IPactBuilder HasPactWith(string providerName);
         IMockProviderService MockService(int port, bool enableSsl = false);
         IMockProviderService MockService(int port, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false);
+        IMockProviderService MockService(int port, bool rewriteLocalhost, JsonSerializerSettings jsonSerializerSettings, bool enableSsl = false);
```
neilcampbell: Can we please remove this, and just add an additional optional param to the end of the above method (after enableSsl = false)? Also can we call `rewriteLocalhost` something like `bindOnAllAdapters`?

PactNet/Mocks/MockHttpService/MockProviderService.cs
```diff
@@ -40,9 +41,13 @@ public class MockProviderService : IMockProviderService
             _httpMethodMapper = httpMethodMapper;
         }
 
-        public MockProviderService(int port, bool enableSsl, string providerName, PactConfig config)
+        public MockProviderService(int port, bool enableSsl, bool rewriteLocalhost, string providerName, PactConfig config)
```
neilcampbell: Can we please rename rewriteLocalhost to bindOnAllAdapters and arrange as the last param?

PactNet/PactBuilder.cs
```diff
@@ -11,10 +11,10 @@ public class PactBuilder : IPactBuilder
     {
         public string ConsumerName { get; private set; }
         public string ProviderName { get; private set; }
-        private readonly Func<int, bool, string, IMockProviderService> _mockProviderServiceFactory;
+        private readonly Func<int, bool, bool, string, IMockProviderService> _mockProviderServiceFactory;
```
neilcampbell: Should also be the last input param

README.md
```diff
@@ -100,6 +100,8 @@ public class ConsumerMyApiPact : IDisposable
 		MockProviderService = PactBuilder.MockService(MockServerPort, true); //By passing true as the second param, you can enabled SSL. This will however require a valid SSL certificate installed and bound with netsh (netsh http add sslcert ipport=0.0.0.0:port certhash=thumbprint appid={app-guid}) on the machine running the test. See https://groups.google.com/forum/#!topic/nancy-web-framework/t75dKyfgzpg
 		//or
 		MockProviderService = PactBuilder.MockService(MockServerPort, new JsonSerializerSettings()); //You can also change the default Json serialization settings using this overload
+		//or
+		MockProviderService = PactBuilder.MockService(MockServerPort, false, new JsonSerializerSettings()); //By passing false as the rewriteLocalhost parameter, the http mock server will only be bound to "localhost", and will not require admin privileges in order to run
```
neilcampbell: Update doco based on code changes

vdex42:
Hi Neil, Thanks for the quick response! 
I have made the suggested changes, please let me know if you have any other further thoughts.
BTW, nice job on the code layout and the unit tests, I found it quite intuitive to work on.

neilcampbell:
Heads up I am playing with a way to remove the urlacl that would have been previously added, so that it's not a breaking change. So yeah it may take a little longer to get this merged in.

neilcampbell:
@vdex42 Worked out a way to handle this. PR has now been merged into master. I will create a new Nuget package tonight when I get home.

vdex42:
Thanks for the speedy resolution. 

I do like your change with the automatic fail-over!

neilcampbell:
@vdex42 Nuget package has been pushed. Can I get you to install and double check everything works correctly for you?

vdex42:
Thanks Neil. I have pulled in the latest nuget package, and it all looks good. 
Cheers.

PR #71, opened 2016-11-15
Update version of dependencies

andy-c-jones:
Due to different (and old) version dependencies @MarkDaviesEsendex and I found it useful to have a version that used the latest versions of pact-net's dependencies. Mainly Newtonsoft.Json and Nancy but some of the other dependencies seemed like pretty low risk to update .

All tests still pass and we have used the package produced by my forks appveyor build in a project that makes use of pact-net in earnest so we believe it is a worthwhile change.

Feel free to ignore this PR if these version updates are unwanted but I thought it was worth offering the changes if you are interested.
The only dependencies we didn't touch were LibLog and xunit because bringing those up to date was a little more involved and didn't provide us with any value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ Build/PactNet*.nupkg
 !packages/repositories.config
 /Build/coverage
 [Ll]ogs
+.vs/
--- a/PactNet.Tests/PactNet.Tests.csproj
+++ b/PactNet.Tests/PactNet.Tests.csproj
@@ -15,7 +15,8 @@
     <TargetFrameworkProfile />
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
     <RestorePackages>true</RestorePackages>
-    <NuGetPackageImportStamp>3ccf767d</NuGetPackageImportStamp>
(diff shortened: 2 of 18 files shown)


andy-c-jones:
I just noticed the PRs don't trigger a CI build. For ease, you can check the appveyor output from our fork [here](https://ci.appveyor.com/project/andy-c-jones/pact-net).

neilcampbell:
@andy-c-jones Thanks for the PR and sorry it has taken so long to respond!
My goal with not updating the dependencies is to make PactNet as compatible as possible with projects using all different dependency versions (including older versions). The versions specified in the nuspec only represent the min version required to use PactNet.

Did you have issues installing PactNet with newer versions of the dependencies?

andy-c-jones:
I believe I did have an issue with newer versions of the dependencies but I can't remember which one off the top of my head, I am going to try and recreate the issue in a sample project.

andy-c-jones:
Either way, given the "min version required" strategy this whole PR would be unsuitable to merge. If I can recreate the problem I will open an issue and/or a new PR.

neilcampbell:
@andy-c-jones Thanks for that! If you can re-create let me know, would be keen to hear about it.

PR #79, opened 2017-01-31
Version 3

cnewham:
Hi there - Our project has a mixed environment (between JS, JVM and .NET) and needed to support pact 3.0 messaging across all apps. Using your awesome project, we branched from version-2 to implement to our current need. At this point, it's more like "3.0ish", but it's a step closer. I hope that you'll consider these changes for the project! @baguaJerry and I would be happy to contribute further.

Summary of Updates:

- Support for messaging in the `PactNet.Models.Messaging` namespace
-- `Message` models the message pact 3.0 structure
-- `PactDslJsonBody` builder is modeled after pact-jvm
-- Heavy use of Newtonsoft.Json and `JToken`
- Added supported matchers (equality, integer, decimal, min, max)
-- Moved the implementations into the `PactNet.Matchers` namespace
-- Took some liberties with a date and timestamp matcher
-- Serialization not up to 3.0 spec (but works with pact-jvm)
- Enhancements to PactBuilder and PactVerifier (`PactMessageBuilder` and `PactMessageVerifier`)
-- More integration with PactBroker  (`PactBrokerClient`)
-- Split out existing interfaces (this is a breaking change - may need some refactoring)
- Enhancements to report output with updates to `MatcherResult` and `MatcherCheck`


uglyog:
Wow, this is awesome.

neilcampbell:
@cnewham Awesome stuff! Will check it out.

neilcampbell:
@cnewham Sorry this is taking a little while to review, this week has been particularly crazy. I'm in transit over the weekend, so will hopefully review on the plane 😄

cnewham:
@neilcampbell No problem! Please let me know if I can help with anything.
Also, we have a couple of additional commits to push, based on some feedback that we're getting while implementing. Coming shortly.

Safe travels!

neilcampbell:
Hi @cnewham,
I just had a look at the PR.

It's starting to look pretty nice, however there is quite a bit left to get v2 and v3 spec compatibility.

Did you do any work towards v2 spec or was it just v3?

I'm thinking that getting v2 spec in before v3 is probably a wise idea.

A few bits of feedback:
- In Consumer.Tests there is a ReceiverPactTest, that should contain a test for the Receiver and ensure the receiver code gets the correct message. At the moment it just exercises the Pact dsl/api. To do this we would need to create some kind of message provider hook. Does the java implementation do this?
- Any reason the models are duplicated in Consumer.Tests project?
- Is there a simpler way to generate objects with matching rules applied? I was thinking something like:

```c#
var body = new
            {
                eventType = new StringType(testEvent.EventType),
                eventId = new GuidType(testEvent.EventId),
                //...
            };
```

or

```c#
var body = new
            {
                eventType = Match.StringType(testEvent.EventType),
                eventId = Match.GuidType(testEvent.EventId),
                //...
            };
```

instead of

```C#
var body = new PactDslJsonRoot()
                    .StringType("eventType", testEvent.EventType)
                    .GuidMatcher("eventId", testEvent.EventId)
                    .DateFormat("timestamp", "O", testEvent.Timestamp)
                    .Object("location")
                        .Object("latitude")
                            .IntegerMatcher("degrees", testEvent.Location.Latitude.Degrees)
                            .IntegerMatcher("minutes", testEvent.Location.Latitude.Minutes)
                            .DecimalMatcher("seconds", testEvent.Location.Latitude.Seconds)
                        .CloseObject()
                        .Object("longitude")
                            .IntegerMatcher("degrees", testEvent.Location.Longitude.Degrees)
                            .IntegerMatcher("minutes", testEvent.Location.Longitude.Minutes)
                            .DecimalMatcher("seconds", testEvent.Location.Longitude.Seconds)
                        .CloseObject()
                    .CloseObject();
```

- There are now two matcher directories, is it possible to have just one for both v2 and v3.
- We need to think about pact file versioning and making sure that things are backwards compatible.
- We need to add doco to the README with usage details once we have formalised the api/dsl.
- I don't think there is really any need for the UsingPactBroker method. You can just detect if the string is http scheme or file scheme. We already do this in PactVerifier.Verify().

I am still quite new to the v3 spec, so I have some questions. These actually may be more for @uglyog 
- With the messaging paradigm, I wanted to the confirm the message publisher is the provider and the message consumer is the consumer?
- Do all the language implementations support v3? How about v2? Is it full support for all the things mentioned in the v2 and v3 Pact specification?
- I'm questioning if it would be easier to move to using the rust implementation and proxy into that. Thoughts?
- Is the java library backwards compatible with a v2 generated pact file?
- I also noticed that the v3 spec mentions jsonpath is not needed, but the example still contains jsonpath. What is the go there?

Hopefully this makes sense, sorry it's a bit of a brain dump :)

Also heads up, I am currently away on holiday, so wont be working on any of this until I get back.

uglyog:
@neilcampbell  Here are some answers:

> With the messaging paradigm, I wanted to the confirm the message publisher is the provider and the message consumer is the consumer?

That is correct, it is the opposite to the HTTP interactions.

> Do all the language implementations support v3? How about v2? Is it full support for all the things mentioned in the v2 and v3 Pact specification?

The JVM and Rust implements support all the spec versions, apart from value generators in V3 (WIP). Ruby supports V2, and the Scala implementation from the ITV guys is V2.

> I'm questioning if it would be easier to move to using the rust implementation and proxy into that. Thoughts?

The Rust implementation is V2 + 90% V3, but the interfaces to it still need to be developed. The Swift consumer library has a POC using it, but the provider side still needs the interface to be worked out.

> Is the java library backwards compatible with a v2 generated pact file?

Yes

> I also noticed that the v3 spec mentions jsonpath is not needed, but the example still contains jsonpath. What is the go there?

We have dropped jsonpath for everything except the bodies. It made no sense to specify `$.header.HEADERA` or `$.path` as a jsonpath expression.

neilcampbell:
Thanks @uglyog!

cnewham:
Hi @neilcampbell, @uglyog - Thanks for the feedback! I should probably start by clarifying that our intent was to add support for messaging with minimal compromise to the existing v1.1. That being said, we completely understand the need for more work to get this fully integrated.

> Did you do any work towards v2 spec or was it just v3?

Aside from implementing most of the v2/v3 matchers (with exception of the `IncludeMatcher`), we didn't take much consideration for v2 updates when adding support for v3 messaging. That's not to say that the `PactDslJsonBody` can't be extended to handle a request or response body the same way for V2.

> In Consumer.Tests there is a ReceiverPactTest, that should contain a test for the Receiver and ensure the receiver code gets the correct message. At the moment it just exercises the Pact dsl/api. To do this we would need to create some kind of message provider hook. Does the java implementation do this?

This test was to prove the builder can build a pact file with the specified message. If I'm understanding this correctly, I think the test that you're referring to is `Provider.Messaging.Tests.ProviderPactTest`, which validates the pact that was generated with the verifier based on the description that is specified.

> Any reason the models are duplicated in Consumer.Tests project?

Not that I'm aware of, we can update this.

>  Is there a simpler way to generate objects with matching rules applied?

First off, I'm completely open to a simpler way to build the body structure. However, I don't generally prefer the usage of anonymous types in .NET. In this particular case, we would lose the state that is required to build the recursive structure of the body, and most importantly, how the matchers are applied to each part during validation.

The `PactMessagingVerifier` does allow anon types and static types to be passed in as the example message to test. Both serialize as expected using `Newtonsoft.Json`

I'll also add that starting with pact-jvm pact builder as an example, it was nice to have a similar look and feel for the pact builder in .NET.

> There are now two matcher directories, is it possible to have just one for both v2 and v3.

Definitely. We moved the matchers into the root namespace, since messaging doesn't use the `MockHttpService`

> We need to think about pact file versioning and making sure that things are backwards compatible.

Agreed. Right now there is nothing that checks the file's pact version metadata, but I think we can utilize this when we attempt to integrate

> We need to add doco to the README with usage details once we have formalised the api/dsl.

Not a problem.

> I don't think there is really any need for the UsingPactBroker method. You can just detect if the string is http scheme or file scheme. We already do this in PactVerifier.Verify().

We can definitely do this. These were additional helper methods that aren't critical

cnewham:
@neilcampbell I may have a different answer regarding the object/matcher builder using dynamic types. I think we can integrate that functionality with the `PactDslJsonBody` structure easier than I had initially thought. @baguaJerry and I will discuss later to see how to make it work.

cnewham:
Summary of additional updates:

- Consolidated Matchers into PactNet.Matchers
- Updates to `PactFile`
-- Consolidated `PactFile` hierarchy into PactNet.Models (Migrated `ProviderServicePactFile` Mocks.Models)
-- Distinguish between `ProviderServicePactFile` and `MessagingPactFile` versions in metadata
- Support for dynamic pact body builder
-- Added `PactDslJsonBody.Parse()` method derived from `ParseBodyMatchingRules`
-- Added overload for `Message.WithBody()` to accept dynamic body structure
- Other various updates based on @neilcampbell  feedback

WIP

- Consolidation of `PactVerifier` to support both V2 and V3 versions of `PactFile`
- Integrate `PactDslJsonBody` structure in `ProviderServiceRequest` and `ProviderServiceResponse` body

Stu-P:
Im trying to update some existing pact tests for a rest service to use the type matchers, like so

```
      .WillRespondWith(new ProviderServiceResponse
      {
          Status = 200,
          Headers = new Dictionary<string, string>
          {
                        { "Content-Type", "application/json; charset=utf-8" },
                        { "access-control-allow-methods", "GET, POST, PUT, DELETE, PATCH, OPTIONS"},
                        { "access-control-allow-headers", "Content-Type, Access-Control-Allow-Headers, Authorization, Accept" }
          },
          Body = new PactDslJsonBody()
                .Int64Type("Request", 895000008)
                .BooleanType("HasError", false)
      });
```

but when my rest client calls the mock pact service and the request matches I dont just get back the example response, but also the matching rules in the response body. 
```
{
	"matchingRules": {
		"$..Request": {
			"match": "type"
		},
		"$..HasError": {
			"match": "type"
		}
	},
	"contents": {
		"Request": 895000008,
		"HasError": false
	}
}
```
 Is that intentional, and if so is there a way to configure the mock service to return only the "contents" in the response body ?

cnewham:
@Stu-P - Serialization with `PactDslJsonBody` inside of `ProviderServiceResponse` and `ProviderServiceRequest` hasn't been implemented yet. It will be exactly as you described:

body will be just the serialized "contents". 

And "matchingRules" will be an additional object serialized as response.matchingRules and request.matchingRules

Blackbaud-AlexKingman:
Hey guys, really looking forward to this update! Thanks for all your work so far :) We've got a lot of internal teams at the company that would GREATLY benefit from this upgrade. It's hard to get the sense of progress on this just from the conversation thread.... is there still momentum here?

neilcampbell:
@Blackbaud-AlexKingman Pact-Net and the Pact model in general is moving to use a shared core, so that the different language implementations can adopt new features etc more quickly. We are currently working on building v2 support using this model see https://github.com/SEEK-Jobs/pact-net/issues/9#issuecomment-306010081. At the moment we are using the Ruby engine as the core library, which does not yet support the v3 spec. At this time I don't have a time frame for v3 support.

PR #85, opened 2017-05-24
Feature/dotnet core no nancy

gruntster:
These changes make PactNet compatible with .NET Core (.NET Standard 1.3) whilst preserving compatibility with .NET 4.0.  Slotting in different HTTP hosts should also be fairly straightforward in the future if need be.  Please let me know what you think.  This PR extends my previous one (targeting .NET Standard 1.6).


bayger:
Are there any plans on merging these changes and releasing a new version of PactNet with .NET Core support?

neilcampbell:
I wont be merging these changes as we are changing the direction of how the core of Pact works to support v2. See https://github.com/SEEK-Jobs/pact-net/issues/9#issuecomment-306010081. I would love some help getting core support on the version-2 branch!

neilcampbell:
@bayger @gruntster ☝️

PR #91, opened 2017-06-23
Fix response body json string handling

kmahon37:
# Overview
Data serialized to json can take a couple different forms depending on whether you have an object, array, string, or number.  These changes fix how "string-like" data is serialized/deserialized in the request/response "body" of a pact interaction when the "Content-Type" header of the body is "application/json".  For example, below is what an interaction request/response "body" would look like before/after this fix for "string-like" data.

### Pact before fix:
`"headers": { "Content-Type": "application/json; charset=utf-8" },`
`"body": "Hello World"`

### Pact after fix:
`"headers": { "Content-Type": "application/json; charset=utf-8" },`
`"body": "\"Hello World\""`

# My Scenario That Surfaced The Issue
I have a .NET backend webservice endpoint that returns a single guid or url (it could also be any other .NET primitive that is convertible to a json string as described in the [Json.Net Serialization Guide](http://www.newtonsoft.com/json/help/html/SerializationGuide.htm)).  When I call that endpoint from my Javascript frontend, I get a simple json response as a double-quoted string.  In my frontend, I create e2e tests that use [Pact JS](https://github.com/pact-foundation/pact-js) to generate a pact file.  In that pact file, I end up with a response body like `"body": "\"https://...\""` because the frontend needs the body to be valid json when it processes the response.  When I tried to run that pact file through my .NET backend webservice using PactNet, the actual would never match the expected result because of how PactNet was stripping away the extra double-quotes that are essential to properly serialize/deserialize json strings.

### DIFF
--- a/PactNet.Tests/Mocks/MockHttpService/Models/HttpBodyContentTests.cs
+++ b/PactNet.Tests/Mocks/MockHttpService/Models/HttpBodyContentTests.cs
@@ -86,7 +86,7 @@ namespace PactNet.Tests.Mocks.MockHttpService.Models
         }
 
         [Fact]
-        public void Ctor1_WithJsonBody_SetsBodyAndContent()
+        public void Ctor1_WithJsonObjectBody_SetsBodyAndContent()
         {
             var body = new
             {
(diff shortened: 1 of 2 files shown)


neilcampbell:
Hi @kmahon37 
Thanks for the nicely detailed PR!
Merging this now.

neilcampbell:
This is now available in Nuget package 1.3.1

kmahon37:
Thanks for the prompt review/merge!

kmahon37:
@neilcampbell Unfortunately, we have identified an issue with my fix in 1.3.1 where the pact verification doesn't work properly.  However, since you recently upgraded the code to support the pact 2.0 spec, I don't have any way to fix the issue.  I've been unable to get your new beta 2.0 package working yet in order to verify it there because of a library we've built around the previous version (and I can't update that library quite yet because of the rest of the company).  Given that I can't fix 1.3.1 anymore, would it be possible to just de-list the 1.3.1 package so that it doesn't cause problems for anyone that may use it?

neilcampbell:
@kmahon37 No problems. I can just cut a fix branch from 1.3.1, so we can make the change.
Nah we can't really remove packages, it just creates more headaches than it solves.

neilcampbell:
@kmahon37 I have just created the 1.3.1-fix branch for you. Keep me posted with the details and how you get on.

neilcampbell:
@kmahon37 Would you like me to roll back the change and release a 1.3.2?

kmahon37:
It is probably best to roll it back for now.   I think it may be a simple fix, but I don't have any time right now to get back into it, and then to fully test it with my different scenarios.  Also, ideally I'd like to test it out with the PactNet 2 beta package and see what it's results would look like (in pact spec 1.1 and 2.0) in the pact file compared to PactNet 1.

All this has also got me re-thinking about whether my idea behind the fix was even right to begin with.  With several different implementations for producing/reading pact files, it is hard to tell which side is right (in my case, did Pact JS create the file incorrectly, or did PactNet read it incorrectly?).

neilcampbell:
No problems will do that.

PR #100, opened 2017-07-15
Add broker support

neilcampbell:


bethesque:
Can you rename `pactVersion` to `consumerVersion` please Neil? It's a subtle difference, but correctly naming that variable will help people understand the API and the underlying model a bit better.

bethesque:
Especially where the interface is exposed to the user.

neilcampbell:
@bethesque Yep sure can, nice catch. The param name was consumerVersion (so that's what is exposed to the user), I just named it badly in the test.

neilcampbell:
And done.

bethesque:
Cheers Neil.

PR #101, opened 2017-07-17
Ported PactNet and PactNet.Tests to netstandard1.5

adamrodger:
The example_description is no longer sent during consumer pact definitions because the previous implementation relied on StackTrace to identify the name of the current test, but this is not available in
.Net Standard 1.5.

The standalone core now copies to `tools` instead of the project directly and also copies to that folder in the NuGet package (which is a bit more standard than the previous location of `build\core`)

I abandoned an idea to use `dotnet pack` instead of `nuget.exe` directly because it was awkward to embed the standalone core in the NuGet this way. There's another issue raised to split this out entirely anyway so it wasn't really worth the effort.

.Net Standard 1.5 is the earliest version because of the dependency on [`System.Runtime.Loader`](https://www.nuget.org/packages/System.Runtime.Loader/) which is needed when registering for an event when the application exits to shutdown the standalone core process. .Net 4.6 is the earliest available version of the full framework because of the dependency on [`System.Diagnostics.Process`](https://www.nuget.org/packages/System.Diagnostics.Process/).


neilcampbell:
Love your work, will review and test it out within the next day or two.

adamrodger:
There's a bug in the NuGet packaging - it doesn't package the `Gemfile` in the standalone core because the match pattern is `**/*.*` and that file has no extension.

I'll fix that tonight.

neilcampbell:
Sorry @adamrodger that I haven't yet reviewed/tested this, it's been a crazy week. I'm hoping to get some time over the weekend to get this merged in.

neilcampbell:
@adamrodger I just notice that Samples are not in the solution anymore. Any reason they have been removed?

adamrodger:
Oops! That was on my Todo list from the original issue and I totally forgot
to do it. I'll add them in this weekend.

neilcampbell:
Awesome, thanks mate!

adamrodger:
@neilcampbell The samples are now added to the solution and both sets of tests are passing.

neilcampbell:
@adamrodger I cant seem to get the tests to run. What are you using to run the tests?

adamrodger:
I was running them either with `dotnet test` on the command line or with
Resharper inside VS2017. What error are you getting?

neilcampbell:
I'm getting System.IO.FileLoadException : Could not load file or assembly 'System.Net.Http'

neilcampbell:
Finally got the tests to run and can see the few failing ones now that you mention. It appears my VS2017 was in a corrupt state and .NET core was not installed correctly. After uninstalling a previously installed version of Web Deployer and Repairing, everything is looking good!

Sounds good, will keep an eye out for the PR update.

adamrodger:
@neilcampbell The tests are all passing now. Notes (included in the commit message):

- InternalsVisibleTo doesn't work in .Net Core with constructors that are both internal and use dynamic parameters for some reason, so these must be public
- PactPublisherTests weren't taking into account the target framework in the file path to the sample pact file
- Integration tests can be a little flaky because the Ruby process doesn't seem to exit properly before the next test starts and so the second test fails to set up the server

I've added a `WaitForExit` when stopping the Ruby process which seems to have helped a bit, but I still had a couple of intermittent failures. If you run the tests again, they typically pass the second time. Not a great answer I know, but it's not something that this PR has caused at least.

adamrodger:
Raised https://github.com/dotnet/coreclr/issues/12988 in relation to the issue with internal constructors with `dynamic` arguments not being accessible even though `InternalsVisibleTo` is enabled.

neilcampbell:
Merged into master now and is available in 2.0.6-beta

Issue #99, opened 2017-07-15
Add support for publishing verification results to the broker


neilcampbell:
Like in https://github.com/pact-foundation/pact-js#publishing-pacts-to-a-broker-and-tagging-pacts

iamdavidfrancis:
Is there a time frame on this feature?

neilcampbell:
@iamdavidfrancis Nah I haven't started this one yet, but we do accept PRs

iamdavidfrancis:
Okay, I can take a look at my schedule and see if I can get one in.

iamdavidfrancis:
I submitted a PR for it. Here are the comments I left on the PR:
> Per Issue #99
> 
> I added two properties on the PactVerifierConfig object and added them to the arguments generated in PactVerifierHostConfig.
> 
> I used the optional parameters from `pact-provider-verifier.bat --help verify`
> 
> ![image](https://user-images.githubusercontent.com/539365/28432620-62dd9230-6d3d-11e7-90aa-c2a316ce8b25.png)

neilcampbell:
Awesome stuff @iamdavidfrancis, will check it all out this weekend.

neilcampbell:
PR has been merged and this feature is available in 2.0.5-beta

PR #102, opened 2017-07-20
Update PactVerifier to allow for publishing results

iamdavidfrancis:
Per Issue #99

I added two properties on the PactVerifierConfig object and added them to the arguments generated in PactVerifierHostConfig.

I used the optional parameters from `pact-provider-verifier.bat --help verify`

image


bethesque:
👍

PR #137, opened 2018-02-20
Isolate the json serializer settings per PactBuilder object

Blackbaud-KrisMahon:
Isolate the json serializer settings per PactBuilder object instead of using global settings.  This allows us to use multiple PactBuilder objects with different serialization settings for mocking multiple services at the same time.

For example:
I have a service (A) that depends on multiple other services (B + C) that I need to setup separate PactBuilder objects for so that I get separate pact mock services.  Service (B) requires special serialization settings in order to work properly, but changing the global serialization settings would also cause service (C) to use those same settings which won't work for service (C).  This change isolates the serialization settings per PactBuilder object so that each pact mock service can have its own serialization settings.

### DIFF
--- a/PactNet.Tests/IntegrationTests/FailureIntegrationTestsMyApiPact.cs
+++ b/PactNet.Tests/IntegrationTests/FailureIntegrationTestsMyApiPact.cs
@@ -16,11 +16,11 @@ namespace PactNet.Tests.IntegrationTests
         {
             var pactConfig = new PactConfig();
 
-            PactBuilder = new PactBuilder((port, enableSsl, consumerName, providerName, host) =>
+            PactBuilder = new PactBuilder((port, enableSsl, consumerName, providerName, host, jsonSerializerSettings) =>
                     new MockProviderService(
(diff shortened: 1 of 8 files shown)


neilcampbell:
Hi @Blackbaud-KrisMahon 
Thanks for the PR. Sounds like a good change to make.

I will review this over the weekend.

neilcampbell:
Merged.

Another option if you have this scenario is to use dynamic/anonymous objects for the req/res body of your interactions and casing the keys the way you want them to serialise. Then there is no need to set any Serialisation settings at all.

Blackbaud-KrisMahon:
Thanks for looking at it and merging it.

I should have mentioned this previously, but we are using the following serialization settings because we are returning some complex xml types (which I'm not a fan of) in one of our services.  I believe we were actually using some dynamic/anonymous objects under PactNet 1.x, but I've been migrating everything to PactNet 2.x and cleaning up the mess we created originally now that we have a better understanding of using pact.

```
ContractResolver = new Newtonsoft.Json.Serialization.DefaultContractResolver
{
    IgnoreSerializableAttribute = false
},
TypeNameHandling = Newtonsoft.Json.TypeNameHandling.Auto
```

PR #140, opened 2018-02-26
Add declaration of ProviderState to ease use of example

ledbutter:
I was following along with the demo and I was initially unable to figure out where `ProviderState` was coming from, this simple update should save future users the trouble I had in looking through the repo to find out where `ProviderState` comes from.


neilcampbell:
Nice catch, thanks for this!

PR #175, opened 2018-12-04
Feature/amqp support

nirpet:


neilcampbell:
@nirpet Awesome stuff, will try get some time over the weekend to check it out!

neilcampbell:
One thing I did noticed was the build for the PR failed https://ci.appveyor.com/project/SEEKJobs/pact-net/builds/20752327

nirpet:
Great,
I found out why the test failed and I fixed the bug.
The build should pass now

neilcampbell:
Tests all pass now 👍

neilcampbell:
@nirpet Just had a quick look and so far it looks great, however still need to sit down and really go through it.

I did notice though that you appear to be using tabs, can you please update to use spaces (1 tab should be 4 spaces, the VS default)?

Also if possible it would be a great help if you can run your eyes over the PR and make sure all the changes were intended, as I noticed a few files with only whitespace changes and a small bit in the readme that I think was accidentally removed.

nirpet:
Sure, I'll revert any unwanted changes.
Also, there is an issue that I forgot to mention in the pull request description.
In the V3 specification, the message pact file also includes a "metadata" section for each message.
It was quite easy to support it on the consumer side, but the pact-verifier does not verify that the provider provides the actual metadata for that message. 
Because of this issue, I've decided to delete the metadata support.
I'm still not quite sure about this decision.. On the one hand, the consumer tests aren't coupled to the provider tests so it is quite weird to delete this code because of a bug in the pact-verifier package. On the other hand, I don't want to confuse the users by allowing to create a section in the pact file that isn't verified. 
What do you think?

nirpet:
I fixed the README file and formatted all of the files. 
I looked at the changes and I think that all of them were on purpose but it is possible that I have missed something.

nirpet:
Thanks!
I'll create an issue.

nirpet:
Created an issue:
https://github.com/pact-foundation/pact-ruby-standalone/issues/26

neilcampbell:
@nirpet Looking at the PR a bit deeper, it looks really good however i'm finding it difficult to give feedback on naming and code organisation without trying a few different options myself, and seeing what feels right.

I'm thinking of taking your branch and merging it onto a new branch in this repo, then we can both collaborate back and forth via PRs. Thoughts?

neilcampbell:
Nah I'll sort it all out.

nirpet:
Hey @neilcampbell, any update?
Also, I found an open issue in the pact-message-ruby package regarding the metadata issue:
https://github.com/pact-foundation/pact-message-ruby/issues/6

neilcampbell:
@nirpet Yep saw that, probably worth adding the metadata stuff back in that way it's there. We can add a comment in the readme later to make it a known issue.

Sorry I have been busy with other stuff the last few days. Hoping to get some more time this week. Basically working on making the samples better to start with, that way we have a nice example and can tweak the DSL based on what feels right in the sample. Hopefully will get a PR to you in the next few days.

nirpet:
Opened a new pull request with the metadata support: https://github.com/pact-foundation/pact-net/pull/176

neilcampbell:
@nirpet Are you able to explain how the message producer verification works? I am trying working on separating the pact http samples from the message pact samples and am a bit confused. Ideally I think we want something like https://github.com/pact-foundation/pact-js#provider-producer. Where I am currently at is on the `update-samples` branch

nirpet:
Yes, I tried following the instructions in the message-pact-demo repo (https://github.com/pact-foundation/pact-message-demo). As a provider you need a controller that wraps the publisher and able to expose a message that was published to the message broker over HTTP (in the samples I created the VereficationController). The MessageInvoker sets up the correct provider states, chooses the publisher method, invokes the publisher and replaces the message broker so it gets the message that is being published instead of the message broker. The MessageInvoker returns the message that has been published and the VereficationController exposes the message over HTTP in the expected pact-provider-verifier contract.

neilcampbell:
I thought it might be doing something like that. Is there any configuration to specify what endpoint pact message is calling? What's the default? Also in a messaging context there may be no http server at all, so im wondering if its better to standup a lighyweight proxy http server as part of pact message. Thought? Im guessing thats what pact js is doing.

nirpet:
Actually, you don't use pact-message to verify the interactions. Pact-provider-verifier package uses the pact-message package internaly, so verifying a contract file containing message interactions is pretty much the same as verifying a contract file containing interactions over HTTP. I think that the endpoint it calls is hardcoded inside the package but you can work around it by providing the service base uri to be the url for your messages verification endpoint (the default is the service base uri). 
Regarding the http server, I also thought about it but I wasn't sure what we should do. I didn't want to add dependencies to the pact-net package so I let the user upload the server by himself like we do in the current solution. If the package would have been separated to multiple packages (pact-net-consumer, pact-net-provider-http, pact-net-provider-amqp) it would defentily be a good idea.

neilcampbell:
Yeah the other thing is that I'm not sure if there is a web server implementation in .net that works cross platform on. net standard/core and full .net.

nirpet:
So we'll leave it as is for now?

nirpet:
Hi @neilcampbell, any update?

neilcampbell:
@nirpet Sorry, life has been very busy at the moment and haven't done much more.
I'm going to try sit down this weekend and come up with a list of the remaining work. 

Hopefully then we have a bit of a plan and can both pick up smaller pieces of work until it's done.

nirpet:
Sure @neilcampbell.
We are using the same package to verify the message pacts (pact-provider-verifier), I guess that failing message pacts will have the same error messages as for failing http pacts but I'll check.

neilcampbell:
@nirpet Created a PR for the readme https://github.com/pact-foundation/pact-net/pull/183

nirpet:
Hi @neilcampbell,
I finally had some time to complete my task.
I did found one strange behavior but it happens also for HTTP pacts.
When the regex validation fails along with the correct error message we get the entire stack trace from the pact-provider-verifier package. (I don't think it can be fixed inside Pact-Net)
Other than that everything looks good, if you approve I'll delete this task from the TODO file.

Gokulaprakash:
Hi, Do you have any update on when this will be merged to master?

neilcampbell:
@nirpet How did you end up setting up the samples to capture any consumer failure messages? I started down the track of IMessageSink, but that didn't work for me. In the message sample in `Handle_WhenAPetIsCreated_SavesThePet` if you change `type = Match.Regex(pet.Type.ToString(), "^(Dog|Cat|Fish)$")` to `type = Match.Regex(PetType.Dog.ToString(), "^(Cat|Fish)$")`, you should see what I mean.

neilcampbell:
@Gokulaprakash No. Heads up this branch actually won't be merged into master, the `message-pact` one will. If you want to use it, you can build your own nuget package from that branch.

taranovskyi:
@neilcampbell there is an existing project for message verifications - https://github.com/Mattersight/pact-net-messages
probably, you can take a look, so it will help to release message functionality quicker

nirpet:
@neilcampbell, maybe I've missed something. 
I'm really busy right now so I'll check this out only next week.

neilcampbell:
@bartschotten I haven't tried running them on Linux (only Windows). It would be awesome if you could look into it!

That reminds me, I'll need to merge master into `message-pact` again.

bartschotten:
When testing with the included sample projects I get the same issue. I've also noticed that, even on Windows, the MessagePact does not generate log files, and that the provided Outputter is never called. This unfortunately makes it very hard to debug.

bartschotten:
Sorry for spamming this thread. I'm just thinking out loud a bit here, but the output I've been able to generate suggests that there's a problem parsing the json on Linux:

`/mnt/c/Repos/Pact-Net/Samples/MessageZooEvents/ZooEventsConsumer.Tests/bin/Debug/netcoreapp2.2/pact-linux-x86_64/lib/vendor/ruby/2.2.0/gems/json-2.1.0/lib/json/common.rb:156:in 'parse': 757: unexpected token at ''{id:{json_class:Pact::SomethingLike,contents:2},name:{json_class:Pact::SomethingLike,contents:Terence},type:{json_class:Pact::SomethingLike,contents:Giraffe}}'' (JSON::ParserError)
        from /mnt/c/Repos/Pact-Net/Samples/MessageZooEvents/ZooEventsConsumer.Tests/bin/Debug/netcoreapp2.2/pact-linux-x86_64/lib/vendor/ruby/2.2.0/gems/json-2.1.0/lib/json/common.rb:156:in 'parse'
        from /mnt/c/Repos/Pact-Net/Samples/MessageZooEvents/ZooEventsConsumer.Tests/bin/Debug/netcoreapp2.2/pact-linux-x86_64/lib/vendor/ruby/2.2.0/gems/json-2.1.0/lib/json/common.rb:335:in 'load'
        from /mnt/c/Repos/Pact-Net/Samples/MessageZooEvents/ZooEventsConsumer.Tests/bin/Debug/netcoreapp2.2/pact-linux-x86_64/lib/vendor/ruby/2.2.0/gems/pact-message-0.5.0/lib/pact/message/cli.rb:24:in 'reify'`

bartschotten:
I've found the issue. The problem occurs when calling the ruby script with a json string as an argument like this (set in the ReifyCommand class):
`reify '{"key":"value"}'`

This works on Windows but on Linux the double quotes get lost somewhere.
It does work when it is called like this:
`reify "{\"key\":\"value\"}"`

Now let's see if I can fix it in a less hacky way than I currently have.

neilcampbell:
@bartschotten Nice work! Also if you could do it on the `message-pact` branch rather than this PR, that would be sweet!

neilcampbell:
@bartschotten Also heads up I just merged the latest `master` into `message-pact`.

bartschotten:
There's another issue that I ran into while running on WSL or on docker with volume sharing. It causes an infamous ruby warning (see for example https://github.com/Microsoft/WSL/issues/1426), which messes up the output that's captured in MessagePact.VerifyConsumer. 
Locally I did an ugly fix to avoid this, but we may want to think about a proper way to handle this (or not).

drkghost:
Hi, @neilcampbell @bartschotten @nirpet 
Please refer to my question https://github.com/pact-foundation/pact-net/issues/246#issuecomment-759106896
I'd like know if I can help with smth, and which branch can be taken for further work

Thank you in advance

adamrodger:
Closed due to #315 being merged instead to add message pact support to  v4.x

PR #176, opened 2018-12-17
Feature/metadata support

nirpet:


neilcampbell:
@nirpet Looks like your VS might be setup for tabs again.

nirpet:
Yeah you're right, don't know why. I'll fix it.

nirpet:
Fixed. It happened because of my resharper settings.

PactNet.Tests/PactMessage/MessagePactTests.cs
```diff
@@ -11,233 +11,239 @@
...
+				new ProviderState
+				{
+					Name = "Test state"
+				}
+			};
+			var expectedContent = new
+			{
+				Test = "Test"
+			};
+			var expectedMetdata = new
```
neilcampbell: This file still appears to be using tabs.

PR #177, opened 2018-12-24
Update samples

neilcampbell:
This PR is to split the message pact samples from the http ones, so that's it's easier to follow.

From here I plan on going through the code and making sure everything sounds correct and is somewhat consistent with other implementations, checking the code lives in the correct place, ensuring a user can easily diagnose failures and writing doco.

@nirpet Are you able to review this?


nirpet:
Yes, I'll try to finish the review in the next few days.

Samples/HttpEventApi/Provider.Api.Web.Tests/EventAPITests.cs
```diff
@@ -1,76 +1,51 @@
...
+                pactVerifier
+                    .ProviderState($"{serviceUri}/provider-states")
+                    .ServiceProvider("Event API", serviceUri)
+                    .HonoursPactWith("Event API Consumer")
+                    .PactUri($"..{Path.DirectorySeparatorChar}..{Path.DirectorySeparatorChar}..{Path.DirectorySeparatorChar}..{Path.DirectorySeparatorChar}Consumer.Tests{Path.DirectorySeparatorChar}pacts{Path.DirectorySeparatorChar}event_api_consumer-event_api.json")
+                    .Verify();
+            }
+        }
+
+        public virtual void Dispose()
```
nirpet: Delete?

Samples/MessageZooEvents/ZooEventsConsumer.Tests/SavePetsTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            var consumer = new SavePets(stubRepo);
+
+            //TODO: Why is this an object?
+            var providerStates = new[]
+            {
+                new ProviderState
+                {
+                    Name = "there is a Pet animal",
+                }
+            };
```
nirpet: In the v3 message pact contract (v3 specification) the provider state was changed a little bit as stated here (https://github.com/pact-foundation/pact-specification/tree/version-3#allow-multiple-provider-states-with-parameters). Actually added a "params" property and deleted it because it wasn't supported. Let's add this property and update the known issue.

Samples/MessageZooEvents/ZooEventsConsumer.Tests/SavePetsTests.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class StubPetRepo : IPetRepo
+    {
+        public bool SavePetWasCalled { get; private set; }
+
+
+        public void SavePet(Pet pet)
+        {
+            SavePetWasCalled = true;
+        }
+    }
```
nirpet: Maybe it will be clearer to use a mocking library

Samples/MessageZooEvents/ZooEventsConsumer/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace ZooEventsConsumer
+{
+    class Program
+    {
+        public static void Main(string[] args)
+        {
+            //TODO: Is it worth showing a Rabbit/AMQP example?
```
nirpet: It would be nice to have. Don't think it's critical though

Samples/MessageZooEvents/ZooEventsProducer.Tests/ZooProducerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            using (var host = builder.Build())
+            {
+                host.Start();
+            
+                //Act / Assert
+                IPactVerifier pactVerifier = new PactVerifier(config);
+                    pactVerifier
+                        .ServiceProvider("Event API", serviceUri)
+                        .HonoursPactWith("Event API Message Consumer")
```
nirpet: How does this work? You've changed the consumer + provider names

Samples/MessageZooEvents/ZooEventsProducer/Program.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace ZooEventsProducer
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            //TODO: Is it worth showing a Rabbit/AMQP example?
+
+            //You're messaging/event technology binding/wiring would go here
+            //which would to register events that are published a corresponding handler
```
nirpet: Please update comment for the producer

Samples/MessageZooEvents/ZooEventsProducer.Tests/ZooProducerTests.cs
neilcampbell: They're actually not used for anything, it's just part of the DSL to make it feel nicer. Will update though.

neilcampbell:
I definitely will add something to the readme about that. Will wait until I write the readme. I'm planning on splitting into a http readme and a message readme and link to both from the main readme

neilcampbell:
@nirpet I just noticed that in the new sample only one consumer interaction (of the two) is being written to the pact json file. Any ideas why?

nirpet:
That's weird. Kind of a dumb question but is there a chance that you were running only one test so it deleted the other interaction?

neilcampbell:
Nah definitely ran them both. Even put a breakpoint in the Dispose to make sure Build was only called once. Were multiple written when you submitted the first PR?

Samples/MessageZooEvents/ZooEventsConsumer.Tests/SavePetsTests.cs
nirpet: Not sure actually, I think that I've tested it and it created multiple interactions with the same description and different provider state. Worth testing.

neilcampbell:
Worked out what is causing it, it's an xunit quirk.

neilcampbell:
@nirpet I just noticed an issue (which I can reproduce in the original PR). When running both of the consumer sample tests together, the second one to runs handler receives both events registered with Pact message.

nirpet:
Nice catch @neilcampbell. I think that I've figured out what the issue is, will send you a pull request soon.

PR #179, opened 2019-01-01
fixed multiple interactions bug

nirpet:
https://github.com/pact-foundation/pact-net/pull/177#issuecomment-450570550


neilcampbell:
@nirpet Nice one, thanks for that!

PR #193, opened 2019-04-28
Easier cross-platform development

mcon:
I've made the following changes which more easily allow for development of pact-net on non-Windows OSs: all of the unit tests pass for me now on Ubuntu, and the CI confirms Windows still works. I haven't added any CI for Ubuntu at this stage.

I plan on making a further PR once I've tidied a few things up, migrating the samples to asp.net core, allowing those to be run on any .net core OS.

Obviously happy to make any changes you'd like.

Changes:
- Platform-specific nuget packages loaded for test project based on OS / arch
- Only target .NET framework if running on Windows
- Allow the "write-to-stderr" and "write-to-stdout" tests to pass on non-Windows OSs
- Update netcoreapp version for tests to 2.0

### DIFF
--- a/PactNet.Tests/PactNet.Tests.csproj
+++ b/PactNet.Tests/PactNet.Tests.csproj
@@ -1,9 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.1</TargetFramework>
-    <TargetFramework Condition=" '$(OS)' == 'Windows_NT' " >
+    <TargetFrameworks>netcoreapp1.1</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' " >
       net46
-    </TargetFramework>
+    </TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup Condition="$(TargetFramework.StartsWith('net4'))">
     <DefineConstants>$(DefineConstants);USE_NET4X</DefineConstants>
--- a/PactNet.Tests/PactNet.Tests.csproj
+++ b/PactNet.Tests/PactNet.Tests.csproj
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' " >
+      netcoreapp2.0
+    </TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">
-      net46
+      net46; netcoreapp2.0
     </TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup Condition="$(TargetFramework.StartsWith('net4'))">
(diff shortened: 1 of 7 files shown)


PactNet.Tests/PactNet.Tests.csproj
```diff
@@ -1,6 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' " >
+      netcoreapp2.0
```
neilcampbell: Any reason to target `netcoreapp2.0` and not `netcoreapp1.1` like we were originally doing?

PactNet.Tests/PactNet.Tests.csproj
```diff
@@ -1,6 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' " >
+      netcoreapp2.0
+    </TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">
+      net46; netcoreapp2.0
```
neilcampbell: Same as above

neilcampbell:
Sorry I was a bit slow to review this, I haven't had a great week.
Awesome work, added a couple of comments.
With the samples PR you mention, I was thinking it might be good to have a seperate .NET core sample. In the `message-pact` branch we have restructured the samples, so it would be good to use this sort of structure as that will go into master soon'ish.

PactNet.Tests/PactNet.Tests.csproj
mcon: It did occur to me that .NET core 2.0 is much more heavily used than 1.1, and so running the tests targeting that would likely be more representative of real world usage - if you're not happy to up-version then I'll change it back to 1.1 - sorry, should really have mentioned this explicitly on the PR.

mcon:
Thanks for making the time (even on a bad week)!

Glad to hear the `message-pact` branch is going into master soon - thanks for all your hard work on it (I'm likely to use it in anger fairly soon).

I'll hold off on any crossplatform changes for tests, and keep that on my local branch for now, waiting until you've merged your changes to master.

It did occur to me that in stead of adding the conditional package reference in the tests, it could be added in the main PactNet project as a dependency there, that way users wouldn't have to add both packages to their tests - would that be a welcome addition to this PR for you?

PactNet.Tests/PactNet.Tests.csproj
neilcampbell: My thought behind this was have the tests target the lowest common denominator. My assumption here is if it passes in 1.1 it should also pass in 2.0.

neilcampbell:
@mcon Sorry not too sure what you mean by adding both packages to their tests? Do you mean `PactNet` and `PactNet-{Platform}` packages? Users only need to install `PactNet-{Platform}` and `PactNet` will be installed as a dependency?

mcon:
@neilcampbell Ahh of course, you're right! I'll just make that .NET core version change then.

PR #195, opened 2019-06-10
Fixed usage of quotes in ruby commands for non-WIndows platforms

bartschotten:


neilcampbell:
Nice one @bartschotten. Will take a peek tonight.

neilcampbell:
@bartschotten Thanks for sorting that one!

PR #204, opened 2019-08-26
Update pact ruby standalone version to 1.69.0

bethesque:


neilcampbell:
Going to also bump to 1.70.0

Issue #209, opened 2019-10-01
Multiple custom headers in PactVerifierConfig


scalancea:
In our project we need several custom headers for authorization on API. They also need to be dynamically generated. In my fork I was able to successfully change the `PactVerifierConfig` to accept a dictionary of custom headers.

Questions:
- Was the one custom header limit imposed on purpose? Was there any limitation imposed by ruby under the hood or windows shell?
- Would such a feature be useful for anyone? Should I merge it back here?

Usage example;
```csharp
var config = new PactVerifierConfig
{
    Outputters = new List<IOutput>
    {
        new XUnitOutput(_output)
    },
    CustomHeaders = new Dictionary<string, string>
    {
        // any headers you wish
        { "x-my-appname", "pactIO" },
        { "x-my-appauth", authHeader }
    }
};

//Act / Assert
IPactVerifier pactVerifier = new PactVerifier(config);
    pactVerifier
        .ServiceProvider("Event API", serviceUri)
        .HonoursPactWith("Event API Consumer")
        .PactUri(pactUri)
        .Verify();
```

neilcampbell:
It was my understand that the ruby core only accepted a single header key value pair, however I could be wrong. Did you manage to get it working with multiple headers?

scalancea:
Yes [it is actually possible](https://github.com/pact-foundation/pact-python/blob/cf5d5bcd64f0537adf9ac187b23b64944f1a2043/pact/verify.py#L54-L60). 
Here is an excerpt of my code (it is not pushed yet, though)
```csharp
foreach (var header in config.CustomHeaders.Where(kv => !string.IsNullOrEmpty(kv.Key) && !string.IsNullOrEmpty(kv.Value)))
{
    builder.Append($" --custom-provider-header \"{header.Key}:{header.Value}\"");
}
```

neilcampbell:
Nice. Yeah that would be awesome if you could send a PR for it!

PR #210, opened 2019-10-01
Enable multiple custom headers to be set through PactVerifierConfig

scalancea:
Enable `PactVerifierConfig` to accept multiple custom headers.
Discussion here #209


neilcampbell:
Thanks for this, and also thinking about backwards compatibility. Merged in, will publish a new version of the nuget package now.

PR #221, opened 2019-10-28
Suppress the inclusion of CSS files in the final package.

### LINKED ISSUE #220: NuGet Package Links A Lot of Superfluous Content
The nuge package introduces a lot of superfluous content into the solution, so if, for example, a developer uses Resharper with the default settings, they will see CSS syntax errors from documentation found in the Ruby gems.

image

This has an impact on the "less than native" experience when working with the .Net pact library when compared to the Java implementation.

Silvenga:
Partially resolves #220.

### DIFF
--- a/Build/linux-x64/PactNet.Linux.x64.nuspec
+++ b/Build/linux-x64/PactNet.Linux.x64.nuspec
@@ -14,7 +14,7 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="..\tools\pact-linux-x86_64\**" target="tools\pact-linux-x86_64\" />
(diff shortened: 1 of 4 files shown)


Silvenga:
Waiting on CI verification.

Silvenga:
I can confirm that the changes suppress the CSS files inclusion in the resulting package:

image

bethesque:
We should be able to exclude those from the source package. I'll have a look.

bethesque:
```
8156173 22 Nov 14:45 pact-1.72.2-osx.tar.gz.before
8142537 22 Nov 14:46 pact-1.72.2-osx.tar.gz.after
```

About 13KB saved, according to my calculations. WIN!!!!

Issue #237, opened 2020-04-07
JSON::ParserError when response body contains a JSON primitive


ddemierre:
I'm working on consumer and provider tests with PactNet, but have consistently run into ParserErrors when the response/request bodies contain primitives. As an example:
```
{
      "description": "A GET request to retrieve a user's ID",
      "providerState": "A user that exists in the System",
      "request": {
        "method": "get",
        "path": "/api/users",
        "query": "username=bill",
      },
      "response": {
        "status": 200,
        "headers": {
          "Content-Type": "application/json; charset=utf-8"
        },
        "body": "c43665eb-b9c8-4787-94b6-e19bd65c9225",
        "matchingRules": {
          "$.body": {
            "match": "type"
          }
        }
      }
    }
```

It was claimed in issue #139 (before it was closed) that strings alone are not valid JSON, but [as of at least Dec 2017 this is untrue](https://tools.ietf.org/html/rfc8259). So it would be nice if this discrepancy could get resolved, or at least if some piece of documentation could clarify which JSON spec Pact expects services to follow. Thanks.

bethesque:
This is supposed to be fixed in the version of JSON that we are using. I couldn't reproduce it in the pure ruby version, however, I could in the pact-ruby-standalone code

https://travis-ci.org/github/pact-foundation/pact-ruby-standalone-e2e-example/builds/672307347

I think it's something to do with the really old version of Ruby that the standalone uses.

I've found a workaround where you wrap the document in an array, parse it, and then take the first element. I'll put out a new version of the pact-ruby-standalone.

mefellows:
Nice work Beth!

mefellows:
Reopening until the standalone is upgraded in this library.

bethesque:
Standalone 1.82.2 released.

bethesque:
My bad, 1.82.3. Travis has just taken a really long time to kick off the build.

neilcampbell:
This will be available in PactNet 2.5.3, which is building as we speak.

ddemierre:
Thanks so much for the fast turnaround! This has fixed the issue for REST tests - is it possible to make the same change in the message-pact branch?

ddemierre:
Thanks very much for the fast turnaround - this has fixed the issue in my provider tests, but I believe I'm still seeing it in my consumer tests? I still get a ParserError for something like the following:
`{
  "description": "A valid id",
  "request": {
    "method": "post",
    "path": "/api/devices",
[...]

PR #238, opened 2020-04-08
feat: update standalone to 1.82.2

bethesque:


bethesque:
For https://github.com/pact-foundation/pact-net/issues/237

PR #255, opened 2020-08-06
Fix: await async calls in tests and README, and always use string as a type

akamud:
I was taking a look at the project, and I found a few places where async calls were not awaited. This is a practice that [can lead](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#avoid-using-taskresult-and-taskwait) to some problems in C#. We should not `.Result` async calls, as that could lead to deadlocks and is considered a bad practice in C#.

I fixed all the instances in tests and README, to better guide people looking at the repository for samples. I also fixed all references that used `String` as a class instead of `string` as a type, as that is also considered a [bad practice](https://blog.paranoidcoding.com/2019/04/08/string-vs-String-is-not-about-style.html).

I couldn't fix all the `.Result` occurrences though, there is a significant design issue that I think you should be aware and take a look. The [MockProviderService](https://github.com/pact-foundation/pact-net/blob/master/PactNet/Mocks/MockHttpService/MockProviderService.cs) makes async calls to configure the Ruby server. As many documentation on async says, [async is viral](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#asynchrony-is-viral), and it should bubble up all the way (reasons mentioned in the link), so with that in mind, all the `_adminHttpClient.SendAdminHttpRequest` calls should be awaited, which then makes `RegisterInteraction` a method that should be async, which in turn dictates that `WillRespondWith` should also be async. Now, that's the big problem, `WillRespondWith` is a fluent method, so making it async breaks all the API that was designed to work as a fluent syntax.
Keeping the code the way it is today is a big problem, though, because people's code could be deadlocking when using Pact-net and they will have a hard time diagnosing it.

With that in mind, please consider a different way of designing the fluent syntax, or at least take a look at ways of mitigating the deadlock issue described [here](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#warning-deadlocks). Also, just so you know, .NET 5, which is set [to be released in November](https://github.com/dotnet/core/blob/master/roadmap.md), will provide [sync methods](https://github.com/dotnet/runtime/pull/34948) for `HttpClient`, so you could just call `httpClient.Send` instead of `SendAsync`. That would fix all your problems in the `MockProviderService` class, since locking the main thread (beucase of the sync methods) shouldn't be a big deal in test suites.

### DIFF
--- a/PactNet.Tests/Fakes/FakeHttpMessageHandler.cs
+++ b/PactNet.Tests/Fakes/FakeHttpMessageHandler.cs
@@ -31,18 +31,15 @@ namespace PactNet.Tests.Fakes
             _requestContentReceived = new List<string>();
         }
 
-        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
             _requestsReceived.Add(request);
             if (request.Content != null)
             {
-                _requestContentReceived.Add(request.Content.ReadAsStringAsync().Result);
+                _requestContentReceived.Add(await request.Content.ReadAsStringAsync());
             }
 
-            var tcs = new TaskCompletionSource<HttpResponseMessage>();
-            tcs.SetResult(ResponseFactory());
-
-            return tcs.Task;
+            return ResponseFactory();
         }
     }
 }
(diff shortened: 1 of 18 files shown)


neilcampbell:
Thanks for that changes. Yep all fair points, which I do know about (the string vs String I'd never actually considered). But yeah as you mention it's tough to fix properly without making breaking changes.

Issue #258, opened 2020-09-07
SSL connect error while using PactNet.Windows 


mayankdotnet:
I am using PactNet.Windows package for Contract testing. Using this package I am able to Create consumer contract tests and publish the same to the broker which have a SSL implemented. But when I we are trying to verify the contract using contract broker url following error is being reported:

**SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed "/bin/Release/netcoreapp3.1/pact-win32/lib/ruby/lib/ruby/2.2.0/net/http.rb:923:in `connect'"**

I have tried using different versions of the library including the latest one for each version I tried I have always received same error.

I have tried verifying and publishing the pact through standalone tool: **https://github.com/pact-foundation/pact-ruby-standalone/releases/tag/v1.88.3**

Also I have added environment variable for **SSL_Cert_File** pointing towards the SSL certificate still getting same error every time.

bethesque:
Because I don't have access to a Windows env, I've been unable to reproduce this issue. You can find some history on it here https://github.com/pact-foundation/pact-ruby-standalone/issues/30

I think the workaround was to put the SSL cert into the right place in the pact ruby standalone dir.

thomas-girotto:
> I think the workaround was to put the SSL cert into the right place in the pact ruby standalone dir.

Well if you mean modifying pact-win32\lib\ruby\lib\ca-bundle.crt that would work but that's not really a solution as it couldn't be applied in a CI context where we restore each time the nuget package, and doing it on each dev's machine (and remembering that you should do it again each time you loose your nuget packages cache) is not really possible.

A better solution that worked for me is to set SSL_CERT_FILE environment variable with a path to my own ca-bundle.crt containing the needed certificates (don't forget to set the whole chain of certificates inside).

But with this solution we still need to set this environment variable on every machine that will run the test project (devs and CI), and have our custom ca-bundle.crt installed somewhere on those machines.

So i'd like to do a PR in which i would give the ruby process the SSL_CERT_FILE environment variable that could be set from dotnet config.
Something like:

```
pactVerifier
  .ServiceProvider("MyService", $"http://localhost:{address.Port}")
  .HonoursPactWith("MyClient")
  .PactUri(
    "https://uri/to/myclient-pact.json", 
    new PactUriOptions 
    { 
      SslCaFilePath = "PathToYourOwnCaBundle" <- proposition of new property
    })
  .Verify();
```
[...]

PR #261, opened 2020-10-05
Feat add ssl cert and http proxy env variable

thomas-girotto:
This PR aims to fix [this issue](https://github.com/pact-foundation/pact-net/issues/258).

The main idea is to be able to set SslCertFilePath in PactUriOptions so that SSL_CERT_FILE environment variable given to ruby process contains the given path instead of the path to the default bundle as of today. It allows us to not depend on ORIG_SSL_CERT_FILE environment variable which is intended to be used to override the default value of SSL_CERT_FILE.

While i was on it, i did the same for HTTP_PROXY environment variable. The idea is the same: don't rely on system's environment variable but add the possibility to give it at runtime so that we don't have to setup build agents and developper's machines.

The main problem i ran into when modifying PactUriOptions class was that we used to be able to only create valid instances of this class by calling one of the two existing constructors (to set basic auth or bearer auth).
With my update, we can now have two more available configuration: SslCertFilePath and HttpProxy, which can exist with or without authentication settings.
Constructors overload were not possible as everything is string, so i found acceptable to allow the creation of an empty instance with a parameterless constructor and then call one or several methods to set the right properties...

I also wondered about creating another configuration class, but i didn't like the idea as all of those properties have the same goal: configuring the HTTP call to fetch the pact.

That led me to finally rename PactUriOptions into PactHttpOptions as those options are more about HTTP call itself than the Uri, and i kept PactUriOptions with its original implementation to avoid breaking changes. 
I thought it would also allow us to handle future cleanup nicely as i was now able to create overloads of PactUri method, one with the new PactHttpOptions and one with the old PactUriOptions marked as Obsolete.

Please let me know if you see any issue or improvement you would like to have in this PR :)


PactNet/PactVerifier.cs
```diff
@@ -94,20 +94,39 @@ public IPactVerifier HonoursPactWith(string consumerName)
             return this;
         }
 
-        public IPactVerifier PactUri(string fileUri, PactUriOptions options = null)
+        public IPactVerifier PactUri(string fileUri, PactUriOptions options)
```
neilcampbell, line 97: This is a breaking change.

PactNet/PactUriOptions.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace PactNet
 {
+    [Obsolete("Please use PactHttpOptions instead")]
```
neilcampbell, line 6: Why the need to introduce `PactHttpOptions` rather than keep using `PactUriOptions`? In my mind it would simplify the change and prevent the breaking changes. It was named `PactUriOptions` in case we needed to add options for consumers reading files from disk rather than via HTTP.

neilcampbell: You could still add your new fluent methods and add a default constructor.

neilcampbell (changes requested):
Thanks for the PR. All looks pretty good, just a couple of thoughts/changes.

PactNet/Core/PactVerifierHostConfig.cs
```diff
@@ -38,7 +43,18 @@ public PactVerifierHostConfig(Uri baseUri, string pactUri, PactBrokerConfig brok
                 { "PACT_INTERACTION_RERUN_COMMAND", "To re-run just this failing interaction, change the verify method to '.Verify(description: \"<PACT_DESCRIPTION>\", providerState: \"<PACT_PROVIDER_STATE>\")'. Please do not check in this change!" }
             };
 
-            if(environment != null)
+            if (!String.IsNullOrEmpty(pactBrokerHttpOptions?.SslCaFilePath))
+            {
+                Environment.Add("SSL_CERT_FILE", pactBrokerHttpOptions.SslCaFilePath);
+            }
+
+            if (!String.IsNullOrEmpty(pactBrokerHttpOptions?.HttpProxy))
```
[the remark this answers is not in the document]
thomas-girotto: Good point, done

PactNet/PactUriOptions.cs
thomas-girotto: Yep, i wondered about this renaming... My rational is:
 - there are no options related to reading the file localaly yet, so PactHttpOptions is more explicit. 
 - If the need for local file reading hasn't come yet, it may never come, so i prefer having a more explicit type right now than a looser one that might make more sense later. 
 - if ever at some point we need some, i'm not a big fan of bag objects that carry a bunch of settings that you only fill partialy depending on the context, and i'd rather create another class like PactLocalFileOptions with an overload in the main method PactUri(...), and maybe throw ArgumentExceptions if you mismatch the type of Uri with the type of options in arguments.

But if you disagree with that, the renaming is a specific commi and, i can revert it.

neilcampbell: Yeah I definitely don't disagree with you points, however:

> if ever at some point we need some, i'm not a big fan of bag objects that carry a bunch of settings that you only fill partialy depending on the context, and i'd rather create another class like PactLocalFileOptions with an overload in the main method PactUri(...), and maybe throw ArgumentExceptions if you mismatch the type of Uri with the type of options in arguments.
Yeah me either, however it never starts as a big bag, it slowly creeps up on you. My thoughts at the moment, is that renaming doesn't really make things much better in my eyes, however it does add a bunch of deprecation, which then makes me think about how we go about deprecating it and how long that needs to live for. I feel like maybe let's just add the fluent methods and keep the same name for now.

thomas-girotto: fair enough :)

thomas-girotto: Done

PR #263, opened 2020-10-12
Upgrade to pact ruby standalone 1 88 14

thomas-girotto:
Hello,

I need to get this new standalone version to include [this change](https://github.com/pact-foundation/pact-ruby/pull/225)


neilcampbell:
Will be available in 2.6.1, which is building as we speak

thomas-girotto:
That was quick :)

bethesque:
Baby napping @neilcampbell? ;)

Issue #245, opened 2020-05-18
Remote Mock Server


AshleyPoole:
Is it possible to configure this to use a remote mock server running elsewhere i.e in Docker?

mefellows:
What's the use case Ashley? 

Technically JS supports this by not by (intentional) design (I know some people do it they way)

AshleyPoole:
@mefellows Thanks for getting back so quickly. We have an issue on Windows where we're not able to run Pact.NET due to long file paths in the bundled Ruby runtime (even with Windows long file paths enabled).
One option we were exploring was to run the mock service ourselves using the Docker image for it and having Pact.Net use that already running mocked server instead to work around the issue.

mefellows:
Thanks Ashley. Makes sense to me.

One of the other good reasons to get us to the Rust implementation!

AshleyPoole:
Oh I hadn't heard about that. Is there a timeline on when we could expect that? Sounds like it would help with our issues! :-)

mefellows:
No timeline, but it is something we're working towards incrementally. Good progress has been made in:

* Pact JS (see v3 branch)
* Pact Swift
* Pact C++

For .NET, it's not a "soon" thing, though, so this improvement would be handy.

bethesque:
> Technically JS supports this but not by (intentional) design (I know some people do it they way)

Ruby has always supported this intentionally. There's a `standalone = true` option in the configuration that stops the code automatically starting up the mock service, and lets you manage it. This could be added to the .Net code.

HugoDoyon:
I proposed a fix in the this [pull request](https://github.com/pact-foundation/pact-net/pull/277). Please merge at your convenience :)

Issue #163, opened 2018-09-26
Build failing for PactNet.Windows Ruby file Path being too long. 


Morshed11218:
Whenever I install Nuget Package Pact.Windows and try to build my solution, I'm getting the file path too long error. I even tried to move the project to smallest path possible (C:\) it still fails with that error: 

Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB3021	Unable to copy file "C:\Users\daudmors\.nuget\packages\pactnet.windows\2.3.3\tools\pact-win32\lib\vendor\ruby\2.2.0\gems\pact-provider-verifier-1.15.0\lib\pact\provider_verifier\underscored_headers_monkeypatch.rb" to "bin\Debug\netcoreapp2.0\pact-win32\lib\vendor\ruby\2.2.0\gems\pact-provider-verifier-1.15.0\lib\pact\provider_verifier\underscored_headers_monkeypatch.rb". The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.	Ipreo.NS.OrderService.BuildingBlocks.Monitoring.UnitTests	C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets	4460

Doesn't make sense, based on where my solution is, the path count is less then 260 characters:  "C:\t\dk-ard2rservice-BuildingBlocks\src\Testing\fdreo.NS.drderrervtce.BuildingBlocks.Testing\bin\Debug\netcoreapp2.0\pact-win32\lib\vendor\ruby\2.2.0\gems\pact-provider-verifier-1.15.0\lib\pact\provider_verifier\underscored_headers_monkeypatch.rb"

neilcampbell:
Unfortunately there is not much we can do at this stage. See https://github.com/pact-foundation/pact-net/issues/160

vikasillumina:
@neilcampbell @bethesque @Morshed11218 I just wanted to update on this, that I tried a lot of suggestions on increasing the windows file path from default 260 characters but none of the settings seems to have made any difference. At this point I am using my mac for my development. But I do have team members who only have windows and this continues to be a problem. While I understand that we can't do much until the refactor that @bethesque mentioned in #160 but unfortunately there is no working workaround either, not at least a straight forward one. :( Just my 2 cents on this!

bethesque:
[...]

PR #277, opened 2021-01-23
Remote mock server service for consumer tests

HugoDoyon:
Add the possibility to use a remote mock server service for the consumer tests. 

For instance a remote mock server could be call to a pact-cli docker container.

To run your consumer tests against a remote server, in PactBuilder class, set the "useRemoteMockService" parameter to true. 

See [readme.md](https://github.com/pact-foundation/pact-net/commit/bec336dbf35c12ef7d889f528c45ee6d2897c8af) for more details 

Fix the following issues:

- Remote Mock Server [Issue #245](https://github.com/pact-foundation/pact-net/issues/245)
- Build failing for PactNet.Windows Ruby file Path being too long. [Issue #163](https://github.com/pact-foundation/pact-net/issues/163)

### DIFF
--- a/PactNet/Mocks/MockHttpService/AdminHttpClient.cs
+++ b/PactNet/Mocks/MockHttpService/AdminHttpClient.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -39,9 +41,9 @@ namespace PactNet.Mocks.MockHttpService
         {
         }
 
-        public async Task SendAdminHttpRequest(HttpVerb method, string path)
+        public async Task SendAdminHttpRequest(HttpVerb method, string path, IDictionary<string, string> headers = null)
         {
-            await SendAdminHttpRequest<object>(method, path, null);
+            await SendAdminHttpRequest<object>(method, path, null, headers:headers);
         }
 
         public async Task SendAdminHttpRequest<T>(HttpVerb method, string path, T requestContent, IDictionary<string, string> headers = null) where T : class
@@ -75,6 +77,7 @@ namespace PactNet.Mocks.MockHttpService
 
             Dispose(request);
             Dispose(response);
+
 
             if (responseStatusCode != HttpStatusCode.OK)
             {
--- a/PactNet/Mocks/MockHttpService/AdminHttpClient.cs
+++ b/PactNet/Mocks/MockHttpService/AdminHttpClient.cs
@@ -78,6 +78,10 @@ namespace PactNet.Mocks.MockHttpService
             Dispose(request);
             Dispose(response);
 
+            if (path == "/pact" && headers != null && headers.ContainsKey("fileNameAndPath"))
+            {
+                File.WriteAllText(headers["fileNameAndPath"], responseContent);
+            }
 
             if (responseStatusCode != HttpStatusCode.OK)
             {
--- a/PactNet/Mocks/MockHttpService/MockProviderService.cs
+++ b/PactNet/Mocks/MockHttpService/MockProviderService.cs
@@ -127,7 +127,7 @@ namespace PactNet.Mocks.MockHttpService
             Async.RunSync(() => _adminHttpClient.SendAdminHttpRequest(HttpVerb.Get, $"{Constants.InteractionsVerificationPath}?example_description={testContext}"));
         }
 
-        public void SendAdminHttpRequest(HttpVerb method, string path)
+        public void SendAdminHttpRequest(HttpVerb method, string path, Dictionary<string, string> headers = null)
         {
             Async.RunSync(() => _adminHttpClient.SendAdminHttpRequest(method, path));
         }
--- a/PactNet/Mocks/MockHttpService/MockProviderService.cs
+++ b/PactNet/Mocks/MockHttpService/MockProviderService.cs
@@ -21,6 +21,8 @@ namespace PactNet.Mocks.MockHttpService
 
         public Uri BaseUri { get; }
 
+        public bool UseRemoteMockService { get; set; } = false;
+
         internal MockProviderService(
             Func<Uri, IHttpHost> hostFactory,
             int port,
@@ -28,7 +30,7 @@ namespace PactNet.Mocks.MockHttpService
             Func<Uri, AdminHttpClient> adminHttpClientFactory)
         {
             _hostFactory = hostFactory;
-            BaseUri = new Uri( $"{(enableSsl ? "https" : "http")}://localhost:{port}");
+            BaseUri = new Uri($"{(enableSsl ? "https" : "http")}://localhost:{port}");
             _adminHttpClient = adminHttpClientFactory(BaseUri);
         }
 
@@ -41,7 +43,7 @@ namespace PactNet.Mocks.MockHttpService
             : this(port, enableSsl, consumerName, providerName, config, ipAddress, jsonSerializerSettings, null, null)
         {
         }
-        
+
         public MockProviderService(int port, bool enableSsl, string consumerName, string providerName, PactConfig config, IPAddress ipAddress, Newtonsoft.Json.JsonSerializerSettings jsonSerializerSettings, string sslCert, string sslKey)
             : this(
                 baseUri => new RubyHttpHost(baseUri, consumerName, providerName, config, ipAddress, sslCert, sslKey),
@@ -97,8 +99,6 @@ namespace PactNet.Mocks.MockHttpService
             return this;
         }
 
-        public bool UseRemoteMockService { get; set; }
-
         public void WillRespondWith(ProviderServiceResponse response)
         {
             if (response == null)
@@ -134,25 +134,41 @@ namespace PactNet.Mocks.MockHttpService
 
         public void Start()
         {
-            StopRunningHost();
+            if (UseRemoteMockService == false)
+            {
+                StopRunningHost();
+                _host = _hostFactory(BaseUri);
+                _host.Start();
+            }
 
-            _host = _hostFactory(BaseUri);
-            _host.Start();
         }
 
         public void Stop()
         {
             ClearAllState();
-            StopRunningHost();
+            if (UseRemoteMockService == false)
+            {
+                StopRunningHost();
+            }
         }
 
         public void ClearInteractions()
         {
-            if (_host != null)
+            if (_host != null && UseRemoteMockService == false)
             {
                 var testContext = BuildTestContext();
                 Async.RunSync(() => _adminHttpClient.SendAdminHttpRequest(HttpVerb.Delete, $"{Constants.InteractionsPath}?example_description={testContext}"));
             }
+            else
+            {
+                ClearAllInteractions();
+            }
+
+        }
+
+        public void ClearAllInteractions()
+        {
+            Async.RunSync(() => _adminHttpClient.SendAdminHttpRequest(HttpVerb.Delete, $"{Constants.InteractionsPath}?"));
         }
 
         private void RegisterInteraction()
--- a/PactNet/Mocks/MockHttpService/MockProviderService.cs
+++ b/PactNet/Mocks/MockHttpService/MockProviderService.cs
@@ -97,6 +97,8 @@ namespace PactNet.Mocks.MockHttpService
             return this;
         }
 
+        public bool UseRemoteMockService { get; set; }
+
         public void WillRespondWith(ProviderServiceResponse response)
         {
             if (response == null)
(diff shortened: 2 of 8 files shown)


bethesque:
I can't comment on the code changes, but the concept is 👍🏽

mefellows:
What Beth says - cool!

charlesericjs:
This addition would be very useful! Changes look legit to me.

bethesque:
I think Neil is a bit busy to review/release right now. Is there anyone else who can do it?

jacekmlynek:
Hi,

I have some thoughts related to code changes can comment on it later on, but that is probably all. It doesn't look like I can merge this to master.

JM.

PactNet/Mocks/MockHttpService/AdminHttpClient.cs
```diff
@@ -76,6 +78,11 @@ public async Task SendAdminHttpRequest(HttpVerb method, string path)
             Dispose(request);
             Dispose(response);
 
+            if (path == "/pact" && headers != null && headers.ContainsKey("fileNameAndPath"))
```
jacekmlynek, line 81: I think this approach my not be the best one for a couple of reasons:

1. Using headers here look like workaround. You pass file path as a header and store pact response content, but is not need to be pass to mock server. Furthermore mock server will ignore it. 
2. Why you decide to store pact file in this class at first place? Is just simple http client class, which IMHO should be responsible for handling http request/response. I can only guess that it is because this method do not return response content. However, won't it be more sensible to return http response or response content from http client instead of affecting it with some pact specific logic?
3. Pact cli, used for stand alone mock server can be run with `--pact-dir`, this is how current implementation works so it may looks a bit less consistent. 
4. You have duplicated `/pact` uri path

From my point of view you have two of option to consider:

1. Change pact-net client in a way it will always be responsible for storing pacts in both cases (attached or stand alone ). Consider mock provider or builder class to be responsible for storing instead of http client. As I have mention above  you will need to return at least http content from http client class.
2. Make stand alone mock server responsible for storing pact file as it is now, However this will probably need to figure out what to do with few PactDir property in `PactBuilder` class.

PactNet/Mocks/MockHttpService/MockProviderService.cs
```diff
@@ -125,32 +127,48 @@ public void VerifyInteractions()
...
                 Async.RunSync(() => _adminHttpClient.SendAdminHttpRequest(HttpVerb.Delete, $"{Constants.InteractionsPath}?example_description={testContext}"));
             }
+            else
+            {
+                ClearAllInteractions();
+            }
+
+        }
+
+        public void ClearAllInteractions()
```
jacekmlynek, line 169: It looks like this method should be private

PactNet/Mocks/MockHttpService/MockProviderService.cs
```diff
@@ -125,32 +127,48 @@ public void VerifyInteractions()
...
+            if (UseRemoteMockService == false)
+            {
+                StopRunningHost();
+            }
         }
 
         public void ClearInteractions()
         {
-            if (_host != null)
+            if (_host != null && UseRemoteMockService == false)
```
jacekmlynek, line 157: If something is a bool it is a little bite more cleaner to not compare this to true or false. So above condition can look like:`if(_host != null && !UseRemoteMockService)`

PactNet/Mocks/MockHttpService/MockProviderService.cs
```diff
@@ -125,32 +127,48 @@ public void VerifyInteractions()
...
+            {
+                ClearAllInteractions();
+            }
+
+        }
+
+        public void ClearAllInteractions()
+        {
+            Async.RunSync(() => _adminHttpClient.SendAdminHttpRequest(HttpVerb.Delete, $"{Constants.InteractionsPath}?"));
         }
```
jacekmlynek, line 172: I have one big suggestion to consider for changes in this class.

Instead of setting additional property in this quite big class, Maybe you could consider creating yet another instance of IMockProviderService and extend a bit concept of `_mockProviderServiceFactory` in PactBuilder so it will not be just Func, but full citizen class? This class based on `useRemoteMockService` param will decide, which realization provide. BTW I do not see why `MockProviderService` is not internal, IMHO it should be.

You can even use some strategy or template method pattern. For instance you will have base abstract class with a few abstract methods related to: Start, Stop, Clear (mostly current implementation of MockProviderService) and two non abstract e.g. AttachedMockProviderService and StandAloneMockProviderService, with abstract method implementations. Hope it is clear.

jacekmlynek (commented):
As I have promised I have left few comments/suggestions. Sorry that it took me a while.

Please let me know if something will not be clear.

PactNet/Mocks/MockHttpService/MockProviderService.cs
HugoDoyon: Fixed

mefellows:
Thanks for the collaboration on this folks 👏 

Are either of you keen to help be a co-maintainer of the project? [Let us know](https://share.hsforms.com/19awx96bnRZaitM4zlnyLtg49tri)!

PactNet/Mocks/MockHttpService/MockProviderService.cs
HugoDoyon: That's a very good suggestion! I am definitely taking good note of this. 

I propose to do this refactor in an upcoming separate pull request (as the current one is already big enough IMO)

PactNet/Mocks/MockHttpService/AdminHttpClient.cs
HugoDoyon: Looking into this :)

HugoDoyon: That a good idea! Much cleaner, I have adjusted the code accordingly, thank you!

jacekmlynek:
Hi @mefellows I am also honored with your offer to become co-maintainer. I do not always have so much time I wish to have, but I am keen to help . Thanks JM.

PactNet/Mocks/MockHttpService/AdminHttpClient.cs
```diff
@@ -78,14 +77,18 @@ public async Task SendAdminHttpRequest(HttpVerb method, string path, IDictionary
...
 
-            if (path == "/pact" && headers != null && headers.ContainsKey("fileNameAndPath"))
+            if (responseStatusCode != HttpStatusCode.OK)
             {
-                File.WriteAllText(headers["fileNameAndPath"], responseContent);
+                throw new PactFailureException(responseContent);
             }
 
-            if (responseStatusCode != HttpStatusCode.OK)
+            if (!string.IsNullOrEmpty(responseContent))
```
jacekmlynek, line 85: Nice to have:

IMHO it a little bit cleaner when you define local variable close to their usage. In your case you define response content at the top and use it almost at the bottom.  Additionally I think (I am not 100% sure) in our case ReadAsStringAsync will never return a null, so you can simplify code a bit. 

```
          var responseContent = string.Empty;

            if (response.Content != null)
            {
                responseContent = await response.Content.ReadAsStringAsync();
            }

            Dispose(request);
            Dispose(response);

            if (responseStatusCode != HttpStatusCode.OK)
            {
                throw new PactFailureException(responseContent);
            }

            return responseContent; //If for some reason you think it can be null you can use ?? - responseContent ?? string.Empty
```

jacekmlynek (commented):
Overall code looks good to me, just one more small suggestion.

mefellows:
Awesome work folks! When you're ready to release this thing, let us know (and maybe also ping us in #pact-net in the slack workspace ([join](http://slack.pact.io)|[visit](http://pact-foundation.slack.com)) and we can work through the details.

PactNet/Mocks/MockHttpService/AdminHttpClient.cs
```diff
@@ -82,22 +80,12 @@ public async Task<string> SendAdminHttpRequest(HttpVerb method, string path, IDi
...
 
-            if (!string.IsNullOrEmpty(responseContent))
-            {
-                return responseContent;
-            }
-            else
-            {
-                return string.Empty;
-            }
+            return !string.IsNullOrEmpty(responseContent) ? responseContent : Empty;
```
jacekmlynek, line 83: Since you have used `using static System.String;`, you may also consider using it in terms of `IsNullOrEmpty`

jacekmlynek:
@HugoDoyon look good to me - really small suggestion from my point of view it can be merged.
@mefellows I am already in the slack.

martincostello:
While investigating why an update to 2.6.2 via dependabot has broken a project of mine, reviewing the [diff between 2.6.1 and 2.6.2](https://github.com/pact-foundation/pact-net/compare/2.6.1..2.6.2) shows that this PR introduced binary and source-breaking compatibility issues by adding new members to the public `IPactBuilder` and `IMockProviderService` interfaces.

This change should have ideally been released as a v3.0.0 to comply with SemVer, been done in a way that didn't cause binary breaking compatibility, or at least have been released as a minor version bump to 2.7 if the major version number is being kept in sync with Pact itself.

I'm still trying to work out what change has actually broken my CI - will open an issue if I get to the bottom if it.

martincostello:
It breaks binary compatibility by amending the public interfaces. Any libraries that maybe extend or consume Pact in some way and are distributed further would be broken due to the interfaces being changed.

Adding optional parameters is a binary-breaking change as the method signature changes from a runtime perspective, even if it isn't neccessarily a source-breaking change. Adding the additional property to `IMockProviderService` is source and binary breaking because existing implementations won't have the `UseRemoteMockService` property implemented.

In my case it didn't break the source as I don't do that in my projects, but from a C#/.NET perspective, these changes are breaking to anyone who may have done that with Pact.

mefellows:
Thanks folks. So to remediate this and prevent further frustrations for others, shall I re-release under `3.0.0` and deprecate the `2.6.2` version?

martincostello:
> I'm still trying to work out what change has actually broken my CI - will open an issue if I get to the bottom if it.

Regarding this, I'm experiencing timeouts in some tests I have where the test process seems to be deadlocking. From viewing the diff further, I think the problem has been introduced by the change at the bottom of this file: https://github.com/pact-foundation/pact-net/compare/2.6.1..2.6.2#diff-8ea699804b2cb88e7361e4f10ad670345b83a1d891e980848c80994d6c535104

`SendAdminHttpRequest()` is an async method that returns a `Task<string>` - in version 2.6.1 it wasn't awaited, so didn't cause too much of an issue. However now with the result being used, it's not awaited and might now have caused some subtle timing change which is causing xunit to deadlock/hang.

https://github.com/pact-foundation/pact-net/blob/ffa6ab841cf602ee5d4644165387e77971393550/PactNet/PactBuilder.cs#L128

Should the call be wrapped in `Async.RunSync()` like some of the other calls in the diff like here?

https://github.com/pact-foundation/pact-net/blob/ffa6ab841cf602ee5d4644165387e77971393550/PactNet/Mocks/MockHttpService/MockProviderService.cs#L171

I think the most likely culprit is `.Result` being used here:

https://github.com/pact-foundation/pact-net/blob/572f04fbf8f5b05f30435e6deaca301ad19269ab/PactNet/Mocks/MockHttpService/MockProviderService.cs#L132

jacekmlynek:
@martincostello as far as I understood binary incompatibility was not your case, you just share some generic thought. Is this right?

@mefellows that sound right to me, but I think it is far to wait for @HugoDoyon as well.

mefellows:
> @mefellows that sound right to me, but I think it is far to wait for @HugoDoyon as well.

NP. I've just unlisted the package until we sort this out to err on the side of caution. I'll be offline probably (getting late here) so will pick any actions up in the morning.

jacekmlynek:
I am not a big expert of async Task in c#, but I took a quick look on the code and unfortunately it looks like a classic example of possible deadlock:

1. MockProviderService calling SendAdminHttpRequest which is an async method
2. SendAdminHttpRequest starts http request by calling _httpClient.SendAsync
3. SendAdminHttpRequest awaits _httpClient.SendAsync and will continue running later.
4. MockProviderService synchronously blocks on the Task returned by SendAdminHttpRequest, which blocks context thread 
5. At some point http request task is ready and want continue, but must wait for context thread, which has been block by calling Result - deadlock

In short term I think we could add to `Async` helper class something like this:
```csharp
public static TResult RunSync<TResult>(Func<Task<TResult>> task)
           => _taskFactory
                .StartNew(task)
                .Unwrap()
                .GetAwaiter()
                .GetResult();
```
And used this instead of Result

In long term we should redesign contract behind  SendAdminHttpRequest IMHO it should not expose asyn methods, unless we will not want to expose public async method too. Hope it is clear. I am just not sure how can reproduce @martincostello problem and test purpose fix.

@HugoDoyon  sorry that I didn't catch that in code review.

mefellows:
Thanks for digging in Jacek. I'm out of my depth here (no C#/.NET knowledge) so will provide a supporting role. It would be great to be able to repro, but also, it sounds like the async call is mishandled anyway. 

@martincostello if the patch described above was added to this PR, could you test if the deadlock occurs?

martincostello:
I'll try and repro a patch some time today.

martincostello:
> I'll try and repro and then check for a fix with a patch some time today.

#293

PR #293, opened 2021-05-12
fix: Resolve deadlock introduced in 2.6.2

martincostello:
* Fix deadlock caused by use of `.Result`. Addresses #277 (comment).
* Fix incorrectly named method.
* Use `Path.Combine()` so that the correct directory separator for the operating system is used.
* Use US English for both tenses of _"initialize"_. Also use `nameof()` for the reference to the method name.
* Use the latest version of Visual Studio in AppVeyor ahead of the release of Visual Studio 2022 later this year which will make 2017 even more out-of-date.
* Adds `netstandard2.0` TFM to the PactNet library.
* Updates the tests to use .NET 3.1 instead of 1.1 which is no longer supported.
* Update NuGet package dependencies in the tests, such as xunit, to their latest versions.
* Fix analyzers warnings from xunit.
* Suppress obsolete warning.

I've validated this fix with the private repo that found the bug in the first place, I'm just working on extracting enough code from its Pact tests to repro it with a unit test in this repo's own test suite.

### DIFF
--- a/PactNet.Tests/Core/MockProviderHostConfigTests.cs
+++ b/PactNet.Tests/Core/MockProviderHostConfigTests.cs
@@ -148,7 +148,7 @@ namespace PactNet.Tests.Core
         {
             var config = GetSubject();
 
-            Assert.Equal(false, config.WaitForExit);
+            Assert.False(config.WaitForExit);
         }
 
         [Fact]
--- a/PactNet.Tests/Core/PactVerifierHostConfigTests.cs
+++ b/PactNet.Tests/Core/PactVerifierHostConfigTests.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using PactNet.Core;
 using Xunit;
 
@@ -221,7 +220,7 @@ namespace PactNet.Tests.Core
         public void Ctor_WhenCalled_SetsWaitForExitToTrue()
         {
             var config = GetSubject();
-            Assert.Equal(true, config.WaitForExit);
+            Assert.True(config.WaitForExit);
         }
 
         [Fact]
@@ -256,7 +255,9 @@ namespace PactNet.Tests.Core
                 providerStateSetupUri: new Uri("http://127.0.0.1/states/"), 
                 verifierConfig: new PactVerifierConfig
                 {
+#pragma warning disable CS0618 // Type or member is obsolete
                     CustomHeader = new KeyValuePair<string, string>("Authorization", "Basic VGVzdA=="),
+#pragma warning restore CS0618 // Type or member is obsolete
                     ProviderVersion = "1.0.0"
                 });
 
@@ -345,7 +346,7 @@ namespace PactNet.Tests.Core
         {
             var config = GetSubject();
 
-            Assert.Equal(null, config.Outputters);
+            Assert.Null(config.Outputters);
         }
 
         private void AssertEnvironmentIsCorrectlySet(IDictionary<string, string> expectedEnv, IDictionary<string, string> actualEnv)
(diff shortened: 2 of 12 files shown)


mefellows:
Thanks so much for this!

> Use US English for both tenses of "initialize". 

😱

martincostello:
> > Use US English for both tenses of "initialize".
> 
> 😱

I can make them both non-US spelling 😉

martincostello:
I've added a test based on the pattern we use in the tests that found the issue, but both locally here and in that private repo I can't actually repro it locally, it only happens when the tests run in our TeamCity continuous integration.

It's not trivial for me set that up to build my fork to validate the fix that way, so I'm at a bit of a loss how to actually make the bug repo locally and/or in AppVeyor.

While I was doing that I noticed that the tests were using an old version of xunit and targeting .NET 1.1 which is no longer supported, so I updated those as well in case that might be a variable that was hiding the problem. That didn't do anything to help repro it locally either, but I've submitted those changes with this PR too anyway as it makes things more up-to-date with the .NET ecosystem.

I've also added a `netstandard2.0` TFM to the PactNet project as that's the recommended baseline to use now.

martincostello:
Once the CI is completed, I'll try out the artifact from the AppVeyor build with our private repo in TeamCity and check the issue is resolved with all these changes.

martincostello:
> Once the CI is completed, I'll try out the artifact from the AppVeyor build with our private repo in TeamCity and check the issue is resolved with all these changes.

Confirmed fixed with our private repo's tests using the packages from the [AppVeyor build](https://ci.appveyor.com/project/PactFoundation/pact-net/builds/39130798/artifacts) for 8cc587ad8ecc49b5d309698254e163ed77c67a7d.

jacekmlynek:
Nice bunch of work @martincostello!

I am quite new in this project, so decision should not be mine, but don't you think that it will a bit better to split this PR for two?:
1. One mainly with my  suggestion on fixing [deadlock](https://github.com/pact-foundation/pact-net/pull/277#issuecomment-839234489)
2. Most of other stuff like upgrading .net core and so on

My point is that it may help us, just in case if we missed something and introduce some less important issue :)

Regarding to problem reproduction/cause. 

> I've added a test based on the pattern we use in the tests that found the issue, but both locally here and in that private repo I can't actually repro it locally, it only happens when the tests run in our TeamCity continuous integration.

>While I was doing that I noticed that the tests were using an old version of xunit and targeting .NET 1.1 which is no longer supported, so I updated those as well in case that might be a variable that was hiding the problem. That didn't do anything to help repro it locally either, but I've submitted those changes with this PR too anyway as it makes things more up-to-date with the .NET ecosystem.

I do not feel that .net core version has something do to with this deadlock. I think that this deadlocks do not appear in each environment, only theses with single thread. I do not know Xunit neither your TeamCity set up (I used to use it but it was like 10 years ago :)),  but if I will have to guess, I will say that your CI may run your test using single thread? Is this possible? However as I said earlier I am not c# async Task expert :).  

One more time thanks for your work.

martincostello:
I happy to split it into separate PRs if necessary.

Our TeamCity isn't explicitly set up to run the tests any differently than locally (it just runs `dotnet test` using the same scripting we use locally). I can only assume it's something to do with the OS (Windows Server compared to Windows 10) or the AWS EC2 instance type our build agents use, such as the number of vCPUs.

jacekmlynek:
@mefellows what is your opinion on splitting this for two?

mefellows:
Let's bring in it - the commits are there so we can always separate those out, and i'm very appreciative of what's been done already.

Going forward, we could add some bots to help enforce commit linting etc. so that we can automate the build process. 

For now, let's fix the issue and get a new release going .

PR #297, opened 2021-06-04
feat!: Large refactor to use Rust code via p/invoke and support v3 request/response pacts

adamrodger:
_This MR is a culmination of my efforts in my fork of PactNet which have been discussed extensively on the Pact Foundation Slack. It is still a work in progress as it stands, but is most of the way to completion. The development should now be moved over to the official PactNet repository so that this can be completed._

This PR is a large refactor of PactNet targeting a 4.0.0 release to replace the Ruby core library with the new Rust core library. This has the main advantage that the mock server and verification process happen in-process instead of spawning and managing an external Ruby process, as well as the fact the new Rust core library will receive more maintenance and features for future Pact specs.

In order to facilitate this, a number of breaking changes had to be made, the main ones of which are:

## Deprecating the OS-specific NuGets

The Ruby implementation is delivered via OS-specific NuGet packages, such as `PactNet.Windows` and `PactNet.Linux.x64`. This can be cumbersome to work with when a project is used cross-platform, such as developing on Windows but running CI in Docker/Linux, as it requires conditional statements in the `csproj` file. Additionally, when using tooling such as Visual Studio and upgrading packages, only the OS-specific section is updated so the versions can come out of sync.

The new Rust implementation uses OS-specific FFI libraries to interact in-process, and all of the OS-specific versions of these libraries are delivered as a single NuGet. The native dependencies are simply unpacked and the correct one used at runtime. This makes PactNet easier to consume and removes the risk of accidentally changing versions between OSs.

## Starting and Stopping the Mock Server

In the Ruby implementation, the mock server is started once in a test fixture setup and then each test configures the interaction(s) it needs, performs the client calls, verifies the interactions and then finally clears up the interactions so that the shared mock server is usable by the next test. The shared mock server is then stopped once all the tests are complete.

In the Rust core library, this is quite different. The Rust mock server must be fully configured with interactions _before_ being started. Once the server is started, any changes you make to it have no effect. This means the previous API of PactNet no longer makes sense - if you follow the same pattern then the mock server never has any interactions configured because when it was started (once, before all the tests) then no interactions were configured, and subsequent interaction configuration has no effect.

Instead, the server must be started and stopped per-test _after_ the interactions have been configured. PactNet provides a nice API for this so that the implementation details are hidden from the user.

## Fluent Builder for Requests and Responses

A number of breaking changes in the Pact v3 spec influenced how the Rust core API is structured (such as supporting multiple provider states and query strings stored as a map).

The Rust core implementation is much more fine-grained than the Ruby core library. For example, adding headers and query parameters is done one-by-one in the Rust implementation whereas in the Ruby external process implementation the entire interaction is set up in a single large POST request to the admin REST endpoint.

The Rust API meant the previous .Net API became unworkable. For example, the `Query` property on the request setup was simply of type `object`, which was fine previously because it would be serialised to JSON and the Ruby server would try to take care of the input. The Rust FFI however requires that each query string parameter is passed one at a time. Given the property is of type `object` and therefore could be anything, this would've required a lot of reflection and processing code in PactNet to try and deal with this fact. This would probably result in breaking changes anyway when inputs that previously worked no longer did.

Instead, it was better to design the PactNet API for correctness so that the user can't express invalid states. The API was moved to a fluent builder pattern for both the request and response so that only valid states can be represented.

## v3 Support

The new PactNet API supports v3 request/response pacts (but not message pacts yet), including some of the breaking changes in the Pact spec. This includes new matchers, new features such as provider states with parameters, and the ability to have multiple values for the same header and query string parameter.

With a mind for future v4 support, additional breaking changes in the fluent builder API were added. For example, when adding the request and response body the method is `WithJsonBody` instead of simply `WithBody`. This will make adding multi-part and binary encoded bodies easier when adopting the v4 spec.

## New API Example

An example of the new API for consumer tests is as follows:

```csharp
[Fact]
public async Task IsAlive_WhenApiIsAlive_ReturnsTrue()
{
    IPactBuilderV2 pact = Pact.V2("Event API Consumer", "Event API", config).UsingNativeBackend();

    // more fine-grained fluent builder pattern for setting up the interaction
    pact
        .UponReceiving("a request to check the api status")
            .WithRequest(HttpMethod.Get, "/stats/status")
        .WillRespond()
            .WithStatus(200)
            .WithJsonBody(new
            {
                alive = true
            });

    using (IPactContext context = pact.Build()) // this starts the server using the above interactions
    {
        var client = new EventsApiClient(context.MockServerUri);

        // client requests must take place within this context because the mock server only lives as long as this block
        var result = await client.IsAlive();

        result.Should().BeTrue();
    } // when disposed, the verification takes place, the pact file is written and the server stopped
}
```

An example of the new verifier API:

```csharp
[Fact]
public void EnsureEventApiHonoursPactWithConsumer()
{
    const string serviceUri = "http://localhost:9222";

    using (WebApp.Start<TestStartup>(serviceUri))
    {
        FileInfo pactPath = new FileInfo("path/to/Event API Consumer-Event API.json");

        //Act / Assert
        IPactVerifier pactVerifier = new PactVerifier(config);
        pactVerifier
            .ServiceProvider("Event API", new Uri(serviceUri))
            .HonoursPactWith("Event API Consumer")
            .FromPactFile(pactPath)
            .WithProviderStateUrl(new Uri($"{serviceUri}/provider-states"))
            .Verify();
    }
}
```


adamrodger:
I know the CI is failing by the way. I couldn't really modify that on my fork to use the new project structure, but it's the first thing I'd look at post-merge

yanncourtel:
The rewrite looks good to me.

A few cosmetic changes we will need to address in the future:
1) Split the interfaces into separate files. Maybe do some folder reorganization, like with functional names, Request / Builder / etc.
2) Use the simpler using declaration (in the tests) introduced in the latest c# version
3) I don't know how you feel about naming convention for the class variable with or without '_' which will allow you not to explicitly have 'this' everywhere.

Otherwise, everything is good to me.

mefellows:
Just post hoc confirming getting this PR in ASAP and working from feature branches into `feature/4.0.0` going forward. Amazing work team!!

zetomatoz:
Great work all! Would you have a rough timeline for a beta v.4 of pact-net?

PR #298, opened 2021-06-04
refactor: Migrate provider sample to ASP.Net Core

adamrodger:
This showcases a more modern sample using ASP.Net Core instead of ASP.Net Legacy and also converts to .Net Core so that we can run the CI fully cross-platform (previous version was .Net Framework so Windows-only)


yanncourtel (approved):
Sounds good to me

PR #300, opened 2021-06-06
chore: Verify checksums of downloaded native dependencies

adamrodger:
Needs to wait for issue to be fixed before merging: https://github.com/pact-foundation/pact-reference/issues/114

### DIFF
--- a/build/download-native-libs.sh
+++ b/build/download-native-libs.sh
@@ -27,13 +27,29 @@ download_native() {
     echo "    Platform: $os/$platform"
     echo "    File: $file.$extension"
     echo "    URL: $url"
+
+    echo -n "    Downloading... "
     curl --silent -L "$url" -o "$path/$file.$extension.gz"
     curl --silent -L "$sha" -o "$path/$file.$extension.gz.sha256"
+    echo "OK"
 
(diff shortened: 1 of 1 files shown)


build/download-native-libs.sh
```diff
@@ -27,13 +27,21 @@ download_native() {
...
+
+    echo -n "    Downloading... "
     curl --silent -L "$url" -o "$path/$file.$extension.gz"
     curl --silent -L "$sha" -o "$path/$file.$extension.gz.sha256"
+    echo "OK"
 
-    # todo: verify the sha256. awkward because the sha256 file contains the wrong file name so needs some processing
-    # sha256sum --check --verify ...
+    echo -n "    Verifying... "
+    sed -Ei "s|../target/artifacts/.+$|$path/$file.$extension.gz|" "$path/$file.$extension.gz.sha256"
```
mefellows, line 37: Argh, we should fix those. I copied the existing ones but that bugs me.

mefellows (approved):

build/download-native-libs.sh
adamrodger: Ha, yeah for sure. Remember to come and fix it here as well if you :grin:

PR #302, opened 2021-06-26
feat!: Change consumer interaction verification API

adamrodger:
The previous implementation using an IDisposable context would swallow exception if the inner interaction call threw. The inner call would fail, but then the Dispose method would be called on the context which itself would throw. The second exception would effectively cause the first to be lost and so you couldn't work out why your inner client call had failed.

Example:

```csharp
this.pact
    .UponReceiving("a request to retrieve all events")
        .Given("there are events with ids 'foo', 'bar'")
        .WithRequest(HttpMethod.Get, "/events")
        .WithHeader("Authorization", $"Bearer {Token}")
    .WillRespond()
        .WithStatus(HttpStatusCode.OK)
        .WithJsonBody(expected);

/*
 * Old API with a disposable context
 */ 
using (IPactContext context = this.pact.Build())
{
    var client = new EventsApiClient(context.MockServerUri, Token);

    // this could throw, e.g. if the request is malformed
    IEnumerable<Event> events = await client.GetAllEvents();

    // this could also throw, e.g. the response is not equivalent to expected
    events.Should().BeEquivalentTo(expected);
} // <-- however, you'd get a PactFailureException here from the Dispose which would hide the inner exception

/*
 * New API with an explicit verify call
 */
await this.pact.VerifyAsync(async ctx =>
{
    var client = new EventsApiClient(ctx.MockServerUri, Token);

    // If this throws then the exception will be raised and the verification will never take place.
    IEnumerable<Event> events = await client.GetAllEvents();

    // if this throws (e.g. because the response was unexpected) then it will be raised also
    events.Should().BeEquivalentTo(expected);
}); // The server will still always be stopped here though so that we don't leak ports
```


src/PactNet.Native/NativePactBuilder.cs
```diff
@@ -63,10 +62,58 @@ internal NativeRequestBuilder UponReceiving(string description)
...
+        /// </summary>
+        /// <param name="interact">Action to perform the real interactions against the mock server</param>
+        /// <exception cref="PactFailureException">Failed to verify the interactions</exception>
+        public void Verify(Action<IConsumerContext> interact)
+        {
+            Uri uri = this.StartMockServer();
+
+            try
+            {
+                interact(new NativeConsumerContext
```
yanncourtel, line 75: I would use interact?.Invoke(...)
Or check the interact action is not null before and throw an ArgumentNullException

src/PactNet.Native/NativePactBuilder.cs
```diff
@@ -63,10 +62,58 @@ internal NativeRequestBuilder UponReceiving(string description)
...
+        /// </summary>
+        /// <param name="interact">Action to perform the real interactions against the mock server</param>
+        /// <exception cref="PactFailureException">Failed to verify the interactions</exception>
+        public async Task VerifyAsync(Func<IConsumerContext, Task> interact)
+        {
+            Uri uri = this.StartMockServer();
+
+            try
+            {
+                await interact(new NativeConsumerContext
```
yanncourtel, line 99: same here

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
```diff
@@ -39,23 +56,147 @@ public void UponReceiving_WhenCalled_CreatesNewInteraction()
...
+        {
+            this.builder.Verify(Success);
+
+            this.mockOutput.Verify(o => o.WriteLine(It.IsAny<string>()), Times.Never);
+        }
+
+        [Fact(Skip = "Enable this when the released FFI library supports getting mock server logs")]
+        public void Verify_Logs_WritesLogsToOutput()
+        {
+            this.mockServer.Setup(s => s.MockServerLogs(this.serverUri.Port)).Returns("some logs");
```
yanncourtel, line 89: I would use a variable expectedLogs to use line 89 and 93.

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
```diff
@@ -39,23 +56,147 @@ public void UponReceiving_WhenCalled_CreatesNewInteraction()
...
+        [Fact]
+        public void Verify_NoMismatches_ShutsDownMockServer()
+        {
+            this.builder.Verify(Success);
+
+            this.mockServer.Verify(s => s.CleanupMockServer(this.serverUri.Port));
+        }
+
+        [Fact]
+        public void Verify_FailedToWritePactFile_ThrowsInvalidOperationException()
```
yanncourtel, line 121: Not sure this test is necessary. You are basically just checking there is no try catch that are absorbing the underlying exception...  If in the future, you want to encapsulate the exception into a more functional one, this test will break.

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
```diff
@@ -39,23 +56,147 @@ public void UponReceiving_WhenCalled_CreatesNewInteraction()
...
+            {
+            }
+
+            this.mockServer.Verify(s => s.WritePactFile(It.IsAny<int>(), It.IsAny<string>(), It.IsAny<bool>()), Times.Never);
+        }
+
+        [Fact]
+        public void Verify_Mismatches_WritesMismatchesToOutput()
+        {
+            this.mockServer.Setup(s => s.MockServerMismatches(this.serverUri.Port)).Returns("some mismatches");
```
yanncourtel, line 152: same thing, I would have a variable expectedLogs I'd use line 152 and 163

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
TimothyJones: I agree - on first skim read, I thought this test was confirming that if the pact failed to write it would throw `InvalidOperationException`, but if I'm reading it correctly, it's just a mock that is throwing that exception?

TimothyJones: Oh, wait - is this here to make sure the exceptions are no longer being swallowed? As in, would this test have failed before this change?

TimothyJones (approved):

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
yanncourtel: That's exactly the case. It's not the mock of the method called but the underlying method. I see why this test is here, but not 100% sure it's necessary to test it when writing the pact.

adamrodger: Yes, this test proves that if the "inner" client code throws an exception then we don't swallow it. Without this, we could catch that exception and not rethrow it, which was the problem that this change is addressing

src/PactNet.Native/NativePactBuilder.cs
adamrodger: Added a null check

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
adamrodger: Introduced local constant for this

tests/PactNet.Native.Tests/NativePactBuilderTests.cs
adamrodger: Introduced local constant for this

src/PactNet.Native/NativePactBuilder.cs
adamrodger: Added null check

PR #304, opened 2021-06-28
feat: add enableIpv6 flag to PactBuilder.cs

garden-party-slug:


greghicks01 (approved):
Notice AppVeyor was not working in this PR. Noted that the changes add enableIpv6 boolean to most signatures in IPactBuilder

greghicks01 (changes requested):
Hey - found out why there is a fail in AppVeyor - you need to fix the tests as they pass in A type for the address and not a bool

garden-party-slug:
> Hey - found out why there is a fail in AppVeyor - you need to fix the tests as they pass in A type for the address and not a bool

Thank you for spotting that!

adamrodger:
Can you squash these commits together and provide a summary of the changes in the commit message, along with following the Conventional Commits pattern?

PactNet/IPactBuilder.cs
```diff
@@ -8,10 +8,10 @@ public interface IPactBuilder
     {
         IPactBuilder ServiceConsumer(string consumerName);
         IPactBuilder HasPactWith(string providerName);
-        IMockProviderService MockService(int port, bool enableSsl = false, IPAddress host = IPAddress.Loopback, string sslCert = null, string sslKey = null, bool useRemoteMockService = false);
+        IMockProviderService MockService(int port, bool enableSsl = false, bool enableIpv6 = true, IPAddress host = IPAddress.Loopback, string sslCert = null, string sslKey = null, bool useRemoteMockService = false);
```
adamrodger, line 11: This is a breaking API change and I don't think we're looking to do another major release before moving to the Rust backend instead of Ruby

adamrodger (changes requested):

PactNet/Mocks/MockHttpService/MockProviderService.cs
```diff
@@ -27,28 +27,30 @@ public class MockProviderService : IMockProviderService
             Func<Uri, IHttpHost> hostFactory,
             int port,
             bool enableSsl,
+            bool enableIpv6,
             Func<Uri, AdminHttpClient> adminHttpClientFactory)
         {
             _hostFactory = hostFactory;
-            BaseUri = new Uri($"{(enableSsl ? "https" : "http")}://localhost:{port}");
+            BaseUri = new Uri($"{(enableSsl ? "https" : "http")}://{(enableIpv6 ? "localhost" : "127.0.0.1")}:{port}");
```
adamrodger, line 34: This is also a breaking change - I'm not sure why ipv6 should always host using the hostname but ipv4 would use the loopback IP? Wouldn't it be more consistent to use `::1` for ipv6 if the user wants to use the IP instead of the hostname alias?

Either way, I think that would be a choice that should be left up to the user. At the moment this is a potentially breaking change for them.

PactNet/IPactBuilder.cs
garden-party-slug: I will arrange to move the enableIpv6 flag to the end, after the useRemoteMockService flag. That should avoid a breaking change.

garden-party-slug:
Closing this pull request. Started fresh with more focus on avoiding breaking changes in https://github.com/pact-foundation/pact-net/pull/306

Issue #301, opened 2021-06-11
Can someone please update the pact verification docs?


bethesque:
These are a bit out of date, and don't show the difference between "verifying a single pact at a URL" (used for webhook flow) and "verifying all the pacts for a provider" (normal provider CI).

https://docs.pact.io/implementation_guides/net/readme#publishing-provider-verification-results-to-a-broker

If someone can also add the two different configurations to a new Pact.NET section is this page, that would also be great https://docs.pact.io/provider/recommended_configuration#verification-triggered-by-provider-change

Thanks!

mefellows:
I'm assuming it's this file source that is needed: https://github.com/pact-foundation/pact-net/blob/master/src/PactNet/Verifier/PactVerifierProvider.cs#L31

PR #305, opened 2021-06-29
Updated the documentation to point the Pact Broker method to the brok…

lmasterman:
…er root URI as the expected relationships do not exist at the pact file itself

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ IPactVerifier pactVerifier = new PactVerifier(config);
 pactVerifier
     .ServiceProvider("Something Api", serviceUri)
     .HonoursPactWith("Consumer")
-    .PactUri("http://pact-broker/pacts/provider/Something%20Api/consumer/Consumer/latest") //NOTE: This must be a pact broker url for this feature to work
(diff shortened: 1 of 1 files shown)


bethesque:
Hm. That doesn't seem right. Perhaps we need to alias that method? When fetching a particular pact, it makes sense to have a method called "PactURI" (as the other libraries do) but when fetching the general pacts for verification from the broker, I would expect to set a "PactBrokerBaseURI" (as the other libraries do). Can we make this happen?

lmasterman:
Yeah, I realised it was incorrect just after submitting the pull request, the PactBroker() method takes a base Uri, my confusion came from muddling two separate documentation sources on setting up the pact-net library with a broker and thinking the full Uri including path was sent to the PactBroker() method, it would perhaps be made clearer if there was a commented or case for utilising the broker method instead of PactUri() here?

bethesque:
There are two use cases - one where you verify a specific pact at a known URL (from the webhook, when a pact changes) and one where you're verifying all the pacts for a provider (when the provider changes).

I've got a request open to make these docs clearer https://github.com/pact-foundation/pact-net/issues/301

Issue #303, opened 2021-06-28
Add 127.0.0.1 as a host option


Perfspec:
See hardcoded mention of localhost below.
https://github.com/pact-foundation/pact-net/blob/c2b8dfeeeebf2312bb37d7c54137802b117720bc/PactNet/Mocks/MockHttpService/MockProviderService.cs#L33

When running this server in a docker container, localhost is not easily available, whereas 127.0.0.1 is available as the loopback address.

Could the ability to choose 127.0.0.1 be added?

garden-party-slug:
see this pull request for ideas: https://github.com/pact-foundation/pact-net/pull/304

PR #306, opened 2021-06-30
feat: add useIpv4LoopbackOnly flag

garden-party-slug:
Goal: Resolution to https://github.com/pact-foundation/pact-net/issues/303
Issue Description: localhost hostname often binds to ipv4 and ipv6 loopback, which can cause bind issues when running mocks in docker containers that don't allowing binding to ipv6 as standard.
Solution Description:
- Update PactNet/Mocks/MockHttpService/MockProviderService.cs
--- edit internal and public constructor with boolean flag that determine BaseUri to use ipv4 loopback or not
- Update PactNet/PactBuilder.cs
--- add a mockProviderServiceFactory2 that accepts an extra boolean
- Create PactNet.Tests/PactBuilderUsingFactory2Tests.cs
--- this is a copy of tests in PactNet.Tests/PactBuilderTests.cs but using mockProviderServiceFactory2
- Update PactNet.Tests/IntegrationTests/XunitIntegrationTests.cs
--- share HttpClient using IClassFixture. see this article on benefits of creating fewer instances of HttpClient https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/
--- use port 4323 instead of 4322 since 4322 is used in IntegrationTestsMyApiPact.cs as well. to prevent possible bind conflict.

### DIFF
--- a/PactNet.Tests/IntegrationTests/XunitIntegrationTests.cs
+++ b/PactNet.Tests/IntegrationTests/XunitIntegrationTests.cs
@@ -57,18 +57,15 @@ namespace PactNet.Tests.IntegrationTests
                 .WillRespondWith(response);
 
             // Act
-            using (var httpClient = CreateHttpClient())
+            using (var httpContent = new StringContent(@"{""left"":1,""right"":1}", Encoding.UTF8, "application/json"))
             {
-                using (var httpContent = new StringContent(@"{""left"":1,""right"":1}", Encoding.UTF8, "application/json"))
+                using (var httpResponse = await Pact.HttpClient.PostAsync("/calculator/add", httpContent))
                 {
-                    using (var httpResponse = await httpClient.PostAsync("/calculator/add", httpContent))
-                    {
-                        Assert.Equal(HttpStatusCode.OK, httpResponse.StatusCode);
+                    Assert.Equal(HttpStatusCode.OK, httpResponse.StatusCode);
 
-                        string json = await httpResponse.Content.ReadAsStringAsync();
+                    string json = await httpResponse.Content.ReadAsStringAsync();
 
-                        Assert.Equal(@"{""result"":2}", json);
-                    }
+                    Assert.Equal(@"{""result"":2}", json);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


garden-party-slug:
I can see that the AppVeyor build failed. I will look into this tomorrow. Any comments welcome.

greghicks01 (changes requested):
there are some tests failing in AppVeyor
[xUnit.net 00:00:08.62]     PactNet.Tests.IntegrationTests.XunitIntegrationTests.Subtract_Numbers [FAIL]
  Failed PactNet.Tests.IntegrationTests.XunitIntegrationTests.Subtract_Numbers [6 s]
  Error Message:
   System.Net.Http.HttpRequestException : An error occurred while sending the request.
---- System.Net.WebException : Unable to connect to the remote server
-------- System.Net.Sockets.SocketException : No connection could be made because the target machine actively refused it 127.0.0.1:4322

See if you can replicate locally

adamrodger:
Unfortunately adding any arguments to the ctor at all is a breaking change, even if they are optional. For example, I've just updated one of our codebases to `3.0.0` from `2.5.x` and now I've getting a `MissingMethodException` because of the new optional arg for supporting remote mock servers:

image

I think it would be a better idea to add additional options to the `IPAddress` enum to allow finer control over the addresses used. There can be a switch statement somewhere that configures the host based upon all the available settings. You could have things like `IPAddress.AllAddressesIPv6` and `IPAddress.LoopbackIPv4Only` etc etc. I was gonna say it could be a `Flags` enum but that would be a breaking change anyway, and also probably awkward to implement.

I'm not sure exactly which options needs to be supported though as I'm not sure what the underlying problem is. I run all of our consumer test suites under Docker and get no issues, so I'd need to see an example of when that causes problems related to IPv6.

garden-party-slug:
> Unfortunately adding any arguments to the ctor at all is a breaking change, even if they are optional. For example, I've just updated one of our codebases to `3.0.0` from `2.5.x` and now I've getting a `MissingMethodException` because of the new optional arg for supporting remote mock servers:
> 
> image
> 
> I think it would be a better idea to add additional options to the `IPAddress` enum to allow finer control over the addresses used. There can be a switch statement somewhere that configures the host based upon all the available settings. You could have things like `IPAddress.AllAddressesIPv6` and `IPAddress.LoopbackIPv4Only` etc etc. I was gonna say it could be a `Flags` enum but that would be a breaking change anyway, and also probably awkward to implement.
> 
> I'm not sure exactly which options needs to be supported though as I'm not sure what the underlying problem is. I run all of our consumer test suites under Docker and get no issues, so I'd need to see an example of when that causes problems related to IPv6.

An example of a docker container that doesn't support ipv6 is mcr.microsoft.com/dotnet/sdk:5.0
See this article for a related issue that shows this docker container doesn't fully support ipv6: https://github.com/dotnet/aspnetcore/issues/29928

In my particular use case, I am getting "Connection to ::1 failed." and "(111) Connection refused" when running a consumer test in the above docker container:
----- Inner Stack Trace #1 (PactNet.PactFailureException) -----
   at PactNet.Mocks.MockHttpService.Host.RubyHttpHost.Start()
   at PactNet.Mocks.MockHttpService.MockProviderService.Start()
   at PactNet.PactBuilder.MockService(Int32 port, JsonSerializerSettings jsonSerializerSettings, Boolean enableSsl, IPAddress host, String sslCert, String sslKey, Boolean useRemoteMockService)
   at PactNet.PactBuilder.MockService(Int32 port, Boolean enableSsl, IPAddress host, String sslCert, String sslKey, Boolean useRemoteMockService)
   at MyCompany.PactTest..ctor(Int32 mockServerPort, String consumerName, String providerName, JsonSerializerSettings jsonSerializerSettings)

I like your solution to use the `IPAddress` enum with a switch statement. I will work on that.

garden-party-slug:
I am running my consumer tests in a kubernetes managed docker container.
Rather than turning off ipv6 in pact-net, it can be turned off there.
https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/

PR #309, opened 2021-07-01
add messaging support

yanncourtel:

On the consumer side:
-> integration of the new interop methods for messaging.
-> add new builder for creating messaging interactions.
-> add extensions methods to make it easy to initiate a message interaction.
-> add simple build method to create a messaging pact file.
-> 

On the provider side:
-> add middleware to act as a messaging proxy server (so far only netstandard 2.0 compatible)
-> add pact message verifying support
-> add scenario builder to emulate the push of a message on the provider side (see sample app: EventImporterProviderTests).

Regarding the tests:
-> add unit tests.
-> add some integration tests.
-> add new sample application for integrated testing with a messaging pact.


src/PactNet.Native/IMessageMockServer.cs
```diff
@@ -0,0 +1,103 @@
+namespace PactNet.Native
+{
+    /// <summary>
+    /// Mock server
+    /// </summary>
+    internal interface IMessageMockServer
```
adamrodger, line 6: Oh - is this the messaging version of the above interface? Were the methods above supposed to be deleted?

src/PactNet/IMessageBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <param name="description">Message description</param>
+        /// <returns>Fluent builder</returns>
+        IPactMessageBuilderV3 ExpectsToReceive(string description);
+
+        /// <summary>
+        /// Add a provider state
+        /// </summary>
+        /// <param name="providerState">Provider state description</param>
+        /// <returns>Fluent builder</returns>
+        IPactMessageBuilderV3 Given(string providerState);
```
adamrodger, line 23: This API allows you to call the methods in an invalid order (e.g. you can call `WithMetadata` before you've called `ExpectsToReceive`. The request/response API separates out the interfaces and uses the type-state pattern so that you _must_ call `UponReceiving` first, and therefore you can't perform invalid sequences of operations.

src/PactNet.Native/MessagePactVerifier.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Set the base path for the message interaction model
+        /// </summary>
+        /// <param name="pactUri">the host uri</param>
+        protected internal override void SetBasePath(Uri pactUri)
+        {
+            verifierArgs.Add("--base-path");
+
+            var messageBasePath = pactUri.AbsolutePath != "/"
```
adamrodger, line 33: Yeah this definitely reinforces the idea that this should be configurable.

src/PactNet.Native/MessageScenarioBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            Scenarios.All ??= new Dictionary<string, Func<dynamic>>();
+        }
+
+        /// <inheritdoc />
+        public IMessageScenarioBuilder WhenReceiving(string description)
+        {
+            try
+            {
+                ValidateDescription(description);
```
adamrodger, line 36: Yeah, mentioned elsewhere in the review but the request/response API means that this is unnecessary because it's designed using the type-state pattern. Instead of having to validate that the user called the API in a valid order, you split the interfaces up into the various states and then add methods to each to transition to the next state.

In the request/response API the states are effectively:

- Uninitialised (an interaction hasn't been defined yet)
- Request (defining the request)
- Response (defining the response)
- Verifying (the interaction is built and can now be verified)

That's why there's separate interfaces for `IRequestBuilder` and `IResponseBuilder`. The only way to get to the response is to first create a request, and then use the `WillRespondWith` method to move to the response configuration state. This prevents the user from doing invalid things, and thus removes the need for this kind of validation.

src/PactNet.Native/Scenarios.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Defines the scenarios static list for messaging support
+    /// </summary>
+    public class Scenarios
+    {
+        /// <summary>
+        /// The available scenarios
+        /// </summary>
+        internal static Dictionary<string, Func<dynamic>> All;
```
adamrodger, line 14: This is a global static field and is not thread safe. The consumer tests (at least for request/response) may run interactions in different threads in parallel, and I think that means you'd have major problems with this approach.

src/PactNet.Native/MessageScenarioBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+            catch (Exception)
+            {
+                ClearScenario();
+                throw;
+            }
+        }
+
+        /// <inheritdoc />
+        public void WillPublishMessage(Func<dynamic> action)
```
adamrodger, line 51: Is there any reason this can't have an overload which just takes a `dynamic` directly instead of requiring a `Func` every time?

src/PactNet.Native/Messaging/MessageMiddleware.cs
```diff
@@ -0,0 +1,91 @@
+#if NETSTANDARD2_0_OR_GREATER
+
+using System.IO;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+
+using Microsoft.AspNetCore.Http;
```
adamrodger, line 8: I think this middleware should be moved to a separate package entirely, because it ties the Native package to ASP.Net Core. We should probably have some kind of `PactNet.AspNetCore` package so that people who were using things other than ASP.Net Core (such as classic ASP.Net or NancyFx) can still use Pact.

I think the fact it has to be wrapped in a .Net Standard 2.0+ guard probably reinforces that even more.

src/PactNet.Native/Messaging/MessageMiddleware.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                await _next(context);
+                return;
+            }
+
+            context.Response.StatusCode = (int)HttpStatusCode.OK;
+
+            string jsonRequestBody = await GetRequestBodyAsync(context);
+
+            var interactionDescription = JsonConvert.DeserializeObject<MessageInteraction>(jsonRequestBody)?.Description;
```
adamrodger, line 49: This uses Json.Net directly, but the default jSON serializer in ASP.Net Core now is `System.Text.Json`. We should probably reuse that just in case there are compatibility issues, and also then we can reuse the default options defined by the user when they create their API.

src/PactNet.Native/Messaging/MessageMiddleware.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            if (string.IsNullOrWhiteSpace(interactionDescription))
+            {
+                //See if raising an exception is better
```
adamrodger, line 53: Yeah I think it's definitely better to throw a 500 error instead of just silently failing.

src/PactNet.Native/Models/NativeMessage.cs
```diff
@@ -0,0 +1,18 @@
+namespace PactNet.Native.Models
+{
+    /// <summary>
+    ///
```
adamrodger, line 4: Needs documentation completing.

src/PactNet.Native/NativeMockServer.cs
```diff
@@ -43,75 +81,81 @@ public string MockServerMismatches(int mockServerPort)
...
         public void WritePactFile(int mockServerPort, string directory, bool overwrite)
+            => VerifyResult(() => Interop.WritePactFile(mockServerPort, directory, overwrite));
+
+        private void VerifyResult(Func<int> func)
         {
-            int result = Interop.WritePactFile(mockServerPort, directory, overwrite);
+            var result = func.Invoke();
 
-            if (result == 0)
+            if (result != 0)
```
adamrodger, line 107: I think this read better before because there was less nesting.

src/PactNet.Native/NativeMockServer.cs
```diff
@@ -168,15 +233,30 @@ private static class Interop
...
+            public static extern IntPtr MessageReify(MessageHandle message);
+
             [DllImport(dllName, EntryPoint = "free_string")]
             public static extern void FreeString(IntPtr s);
         }
     }
 
     [StructLayout(LayoutKind.Sequential)]
-    internal readonly struct PactHandle
+    internal struct PactHandle
```
adamrodger, line 251: These should be `readonly` - it's dangerous to change them after you've created them when doing the interop.

src/PactNet.Native/NativePactMessageBuilder.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// <param name="server">Mock server</param>
+        /// <param name="pact">the pact handle</param>
+        /// <param name="config">the pact configuration</param>
+        /// <param name="defaultSettings">Default JSON serializer settings</param>
+        internal NativePactMessageBuilder(IMessageMockServer server, MessagePactHandle pact, PactConfig config, JsonSerializerSettings defaultSettings = null)
+        {
+            this.pact = pact;
+            this.config = config ?? throw new ArgumentNullException(nameof(config));
+            this.server = server ?? throw new ArgumentNullException(nameof(server));
+            message = server.NewMessage(pact, "default message");
```
adamrodger, line 34: This call shouldn't be done in the constructor because it can fail.

src/PactNet.Native/NativePactMessageBuilder.cs
```diff
@@ -0,0 +1,168 @@
...
+        internal void Build()
+        {
+            server.WriteMessagePactFile(pact, config.PactDir, true);
+        }
+
+        /// <summary>
+        /// Verify a message is read and handled correctly
+        /// </summary>
+        /// <param name="handler">The method using the message</param>
+        internal void Verify<T>(Action<T> handler)
```
adamrodger, line 149: Probabably needs an async version also to match the request/response verification.

src/PactNet.Native/NativePactMessageBuilder.cs
```diff
@@ -0,0 +1,168 @@
...
+        private readonly JsonSerializerSettings defaultSettings;
+
+        /// <summary>
+        /// Initialises a new instance of the <see cref="NativePactMessageBuilder"/> class.
+        /// </summary>
+        /// <param name="server">Mock server</param>
+        /// <param name="pact">the pact handle</param>
+        /// <param name="config">the pact configuration</param>
+        /// <param name="defaultSettings">Default JSON serializer settings</param>
+        internal NativePactMessageBuilder(IMessageMockServer server, MessagePactHandle pact, PactConfig config, JsonSerializerSettings defaultSettings = null)
```
adamrodger, line 29: The JSON settings should be removed because we should be using the settings defined on the `PactConfig`. The extension method that constructs this class doesn't actually pass them in, so these will always be `null` in real usage.

src/PactNet.Native/PactExtensions.cs
```diff
@@ -49,6 +49,25 @@ public static IPactBuilderV3 UsingNativeBackend(this IPactV3 pact, int? port = n
...
+        /// Establish a new pact using the native backend
+        /// </summary>
+        /// <param name="pact">Pact details</param>
+        /// <returns>Pact builder</returns>
+        /// <remarks>
+        /// If multiple mock servers are started at the same time, you must make sure you don't supply the same port twice.
+        /// It is advised that the port is not specified whenever possible to allow PactNet to allocate a port dynamically
+        /// and ensure there are no port clashes
+        /// </remarks>
+        public static IPactMessageBuilderV3 UsingNativeBackendForMessage(this IPactV3 pact)
```
adamrodger, line 62: I wonder if it's worth defining an `IMessagePact` and `IMessagePactV3` etc so that the extension method is a bit neater. Then you could have:

```csharp
public static IPactMessageBuilderV3 UsingNativeBackend(this IMessagePactV3 pact)
```

which would means there wouldn't be an API difference between message and non-message pacts.

src/PactNet.Native/PactExtensions.cs
```diff
@@ -49,6 +49,25 @@ public static IPactBuilderV3 UsingNativeBackend(this IPactV3 pact, int? port = n
             return builder;
         }
 
+        /// <summary>
+        /// Establish a new pact using the native backend
+        /// </summary>
+        /// <param name="pact">Pact details</param>
+        /// <returns>Pact builder</returns>
+        /// <remarks>
+        /// If multiple mock servers are started at the same time, you must make sure you don't supply the same port twice.
```
adamrodger, line 58: Is this comment accurate? Message pacts don't create mock servers do they?

src/PactNet.Native/PactNet.Native.csproj
```diff
@@ -107,4 +107,10 @@
     <ProjectReference Include="..\PactNet\PactNet.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Http">
```
adamrodger, line 111: Is this NuGet package still maintained? It looks very old and hasn't been updated for 2.5 years.

src/PactNet.Native/PactVerifier.cs
```diff
@@ -12,7 +12,7 @@ public class PactVerifier : IPactVerifier
     {
         private readonly IVerifierProvider verifier;
         private readonly PactVerifierConfig config;
-        private readonly IList<string> verifierArgs = new List<string>();
+        protected internal readonly IList<string> verifierArgs = new List<string>();
```
adamrodger, line 15: I think it's better to add the necessary extra method(s) onto the verifier interface instead of using inheritance to customise the implementation.

src/PactNet.Native/PactVerifier.cs
```diff
@@ -48,20 +48,31 @@ internal PactVerifier(IVerifierProvider verifier, PactVerifierConfig config)
         /// <returns>Fluent builder</returns>
         public IPactVerifier ServiceProvider(string providerName, Uri pactUri)
         {
-            this.verifierArgs.Add("--provider-name");
-            this.verifierArgs.Add(providerName);
-            this.verifierArgs.Add("--hostname");
-            this.verifierArgs.Add(pactUri.Host);
-            this.verifierArgs.Add("--port");
-            this.verifierArgs.Add(pactUri.Port.ToString());
+            verifierArgs.Add("--provider-name");
```
adamrodger, line 51: I think we should put all the `this.` back on to remove noise in the diff. I almost didn't spot the new `SetBasePath` call below because of those.

src/PactNet/IRequestBuilder.cs
```diff
@@ -88,7 +89,7 @@ public interface IRequestBuilderV3
         /// </summary>
         /// <param name="providerState">Provider state description</param>
         /// <param name="parameters">Provider state parameters</param>
-        /// <returns>Flient builder</returns>
+        /// <returns>Fluent builder</returns>
```
adamrodger, line 92: Ha, good spot!

src/PactNet/Pact.cs
```diff
@@ -53,9 +53,9 @@ private Pact(string consumer, string provider, PactConfig config)
                 throw new ArgumentException("Please provide a valid provider name", nameof(provider));
             }
 
-            this.Consumer = consumer;
-            this.Provider = provider;
-            this.Config = config;
+            Consumer = consumer;
```
adamrodger, line 56: Worth reverting to remove noise from the change

src/PactNet/PactNet.csproj
```diff
@@ -51,4 +51,9 @@
     <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
     <PackageReference Include="System.Threading.Thread" Version="4.3.0" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions">
```
adamrodger, line 55: Similarly to the other ASP.Net Core reference - I don't think these references should be added to the main package. They tie us to a single web framework instead of allowing the user to provide their own.

tests/PactNet.Native.Tests/Messaging/MessageMiddlewareTests.cs
```diff
@@ -0,0 +1,99 @@
+//#if NET5_0_OR_GREATER
```
adamrodger, line 1: These tests are commented out - not sure why.

tests/PactNet.Tests/PactNet.Tests.csproj
```diff
@@ -37,4 +37,9 @@
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net50'">
```
adamrodger, line 40: This import should be unnecessary because the package will be pulled in transitively from the referenced main projects.

adamrodger (changes requested):

samples/EventApi/Consumer.Tests/EventsApiConsumerTestsV3.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    EventId = Guid.Parse("3E83A96B-2A0C-49B1-9959-26DF23F83AEB"),
+                    Timestamp = DateTime.Parse("2014-06-30T01:38:00.8518952"),
+                    EventType = "SearchView"
+                }
+            };
+
+            pact
+                .UponReceiving("a request to retrieve all events with multiple provider states")
+                .Given("there are events with id '45D80D13-D5A2-48D7-8353-CBB4C0EAABF5'")
```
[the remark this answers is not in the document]
yanncourtel: You are right, I wanted to illustrate we could write different instructions but the way it'd be used is for instance to mock another API or component that has nothing to do with the events.

How about something as simple as this :
                .UponReceiving("a request to retrieve all events with multiple provider states")
                .Given("there are events with id '45D80D13-D5A2-48D7-8353-CBB4C0EAABF5'")
                .Given("another provider state")
                
Otherwise, we could add a component in the API sample calling an external API and use the provider state to mock it (overkill?)

samples/EventImporter/Consumer.Tests/EventImporterTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            var worker = new EventsWorker(new FakeEventProcessor());
+
+            _pactMessage
+                .ExpectsToReceive("receiving events from the queue")
+                .Given("A list of events is pushed to the queue")
+                .WithMetadata("key", "valueKey")
+                .WithContent(Match.Type(expected))
+                .Verify<List<Event>>(events => worker.ProcessMessages(events));
+
+            _pactMessage.Build();
```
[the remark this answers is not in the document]
yanncourtel: That's how the old library was doing and we had a encapsulate this statement into a fixture class so agree, not a good idea.
We could call the Build after the Verify (internally in the builder), that way we force the user to verify to write the pact.

samples/EventImporter/Provider/Domain/IEventProducer.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+using Provider.Domain.Models;
+
+namespace Provider.Domain
+{
+    public interface IEventProducer
+    {
+        void SendAsync(IReadOnlyCollection<Event> events);
```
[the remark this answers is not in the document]
yanncourtel: A copy paste mistake

samples/EventImporter/Consumer/DateTimeFactory.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Consumer
+{
+    public class DateTimeFactory
```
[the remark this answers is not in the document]
yanncourtel: Removed

src/PactNet.Native/Constants.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// Defines the constants for pact native 
+    /// </summary>
+    public static class Constants
+    {
+        /// <summary>
+        /// The base path for the pact message
+        /// </summary>
+        public const string PactMessagesPath = "/pact-messages";
```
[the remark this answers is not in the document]
yanncourtel: I'll see where I can put it

src/PactNet.Native/IHttpMockServer.cs
```diff
@@ -1,11 +1,11 @@
-﻿using System;
+using System;
 
 namespace PactNet.Native
 {
     /// <summary>
     /// Mock server
     /// </summary>
-    internal interface IMockServer
+    internal interface IHttpMockServer
```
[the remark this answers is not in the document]
yanncourtel: I can leave it to IMockServer for the default one (http and non-http) and add the IMessageMockServer then.

src/PactNet.Native/IHttpMockServer.cs
```diff
@@ -154,5 +154,92 @@ internal interface IMockServer
...
+        /// <returns></returns>
+        public MessageHandle NewMessage(MessagePactHandle pact, string description);
+
+        /// <summary>
+        /// Set the description of the message description
+        /// </summary>
+        /// <param name="message">message</param>
+        /// <param name="description">message description</param>
+        /// <returns>Success</returns>
+        public bool MessageExpectsToReceive(MessageHandle message, string description);
```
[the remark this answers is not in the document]
yanncourtel: I'll keep them internal

src/PactNet.Native/IHttpMockServer.cs
```diff
@@ -154,5 +154,92 @@ internal interface IMockServer
...
         bool WithResponseBody(InteractionHandle interaction, string contentType, string body);
+
+        /// <summary>
+        /// Write the pact message file
+        /// </summary>
+        /// <param name="pact">the pact</param>
+        /// <param name="directory">the output folder</param>
+        /// <param name="overwrite">overwrite</param>
+        /// <returns></returns>
+        public void WriteMessagePactFile(MessagePactHandle pact, string directory, bool overwrite);
```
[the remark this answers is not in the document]
yanncourtel: hm, that's true. This is going to be a pain in the ass with the current implementation. I'll split the builders and see where to go from here.

src/PactNet.Native/IMessageMockServer.cs
yanncourtel: Exactly. Each interaction model will add their own set of methods

src/PactNet/IMessageBuilder.cs
yanncourtel: As discussed, this will be covered with the splitting of the builders.

src/PactNet.Native/Scenarios.cs
yanncourtel: I forgot the new Dictionary<string, Func<dynamic>>(); which should be thread-safe per microsoft.
The best would be to test it, improve the sample app to add two classes with diff provider tests. I'll see to add it.

src/PactNet.Native/MessageScenarioBuilder.cs
yanncourtel: No, you need a function because you need to test the return of an algorithm and not just an object. In PACT messaging, the most important thing is not to test that an object is the same type as the one in the contract but that the object _pushed in the queue_ is the same. Therefore you would need a way to simulate the pushing of an object.

src/PactNet.Native/Messaging/MessageMiddleware.cs
yanncourtel: Good remark. I'll see to add it. Later we could add any number of package for the versions of the middlewares.

src/PactNet.Native/Messaging/MessageMiddleware.cs
yanncourtel: Not sure, it's the case. Microsoft is going to package newtonsoft in their future framework because system.text hasn't caught up in time... That said, I don't see why this middleware shouldn't work with System.Text.Json.

src/PactNet.Native/NativeMockServer.cs
yanncourtel: Yes, the rule of thump is to nest the non-optimal case, baring is doesn't complicate your code, in order to clearly see what the method is doing optimally :)

src/PactNet.Native/PactExtensions.cs
yanncourtel: Let me see...
We would them call it like that :
IMessagePactV3 pactMessageV3 = Pact.V3Message("Event API Consumer V3 Message", "Event API V3 Message", _config);
_pactMessage = pactMessageV3.UsingNativeBackendForMessage();

And the Pact class would then implement the new IMessagePactV3 ?

In that case, not sure we need to have an IMessagePact interface. All interaction model should implement the IPact interface, no?

src/PactNet.Native/PactExtensions.cs
yanncourtel: copy paste ^^

src/PactNet.Native/PactNet.Native.csproj
yanncourtel: Not sure about it

tests/PactNet.Native.Tests/Messaging/MessageMiddlewareTests.cs
yanncourtel: I have to put them back

mefellows:
Just making a comment here, but it's not immediately relevant (I don't think).

We recently discovered that if a message contract _does_ have metadata in it, the verification will fail unless the provider sends the valid metadata. This was not currently supported via the FFI/CLI.

As of https://github.com/pact-foundation/pact-reference/pull/120 it now is. This [doc](https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_verifier_cli/README.md#verifying-message-pacts) shows how to do it at the framework level. That is, how to pass the metadata to the verifier to include it in verification.

It need not be in this PR, but probably worth thinking about what the message verification API looks like to support this. This is how the JVM supports it: https://docs.pact.io/implementation_guides/jvm/provider/junit/#example-of-message-test-that-verifies-metadata.

In Typescript, we plan on allowing users to return one of two types in a verification - one that just returns the message, and another that allows metadata to be included also.

samples/EventApi/Consumer.Tests/EventsApiConsumerTestsV3.cs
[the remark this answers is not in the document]
yanncourtel: Did it like this to illustrate the point of having multiple provider states:

            pact
                .UponReceiving("a request to retrieve all events with multiple provider states")
                .Given("several events have already been created")      //Can create events on the provider end
                .Given("all events have been processed")                //Can set other event information such as execution date and status on the provider end
                .WithRequest(HttpMethod.Get, "/events")
                .WithHeader("Authorization", $"Bearer {Token}")
                .WillRespond()
                .WithStatus(HttpStatusCode.OK)
                .WithJsonBody(expected);

samples/EventImporter/Consumer.Tests/EventImporterTests.cs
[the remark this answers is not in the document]
yanncourtel: I stuffed it into the Verify Method. Maybe, by the end of the PR I might put it async and call it from a PactMessageBuilder class, we will see :)

samples/EventImporter/Provider/Startup.cs
```diff
@@ -0,0 +1,91 @@
...
+                app.UseSwagger();
+                app.UseSwaggerUI(c => c.SwaggerEndpoint("/swagger/v1/swagger.json", "Provider v1"));
+            }
+
+            app.UseRouting();
+
+            app.UseAuthentication();
+            app.UseAuthorization();
+
+            //Test the message with the PACT VERIFIER CLI
```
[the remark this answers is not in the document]
yanncourtel: I definitively need to remove this now since I use a Scenario system in the test project :)

src/PactNet.Native/IMessageMockServer.cs
yanncourtel: Can't remember what this comment was referring to but I segregated the interfaces for messaging and non messaging so all the methods that are supposed to be for messaging are carried by this interface. They were duplicated in the IMockServer interface so I removed them.

src/PactNet.Native/NativePactMessageBuilder.cs
yanncourtel: With the builder state pattern, it is now done by the ExpectedToReceive before it creates the message via another builder.

src/PactNet/PactNet.csproj
yanncourtel: I separated the middleware into its own project. A couple of things:
- We will need to think of a alternate way of invoking scenario for other .net framework version. (honestly, going with a message controller is not a good idea ^^)
- We could then add the provider middleware into this new project as well to be re-used.

src/PactNet.Native/Messaging/MessageMiddleware.cs
yanncourtel: I'm not sure how to handle exceptions on middleware. I'll research and will add it on a later PR

src/PactNet.Native/PactExtensions.cs
yanncourtel: I tried to add a IMessagePact and IPactMessageV3 with a UsingNativeBackend()... 
More duplicated code but let's see what we prefer.

PR #310, opened 2021-07-06
feat: add mock server logs to consumer test output

adamrodger:
Upgrade the FFI libs to latest, and enable retrieving mock server logs now that they're supported in this version.

Note: The logs themselves still don't work because there's no way to enable them via FFI at the moment, but once the FFI libs enable logs internally then this will start working.


mefellows:
I haven't tested this myself yet.

But, I think it doesn't work because you don't initialise logging via `log_to_buffer`. This is not very obvious (i.e. not documented in the FFI README but is noted in the implementation rust docs):

```rust
/// Fetch the logs for the mock server. This needs the memory buffer log sink to be setup before
/// the mock server is started. Returned string will be freed with the `cleanup_mock_server`
/// function call.
///
/// Will return a NULL pointer if the logs for the mock server can not be retrieved.
#[no_mangle]
pub extern fn mock_server_logs(mock_server_port: i32) -> *const c_char {
```

The is alluding to calling the `log_to_buffer` option, which would replace `init` (it takes the same argument).

adamrodger:
Yeah the problem is that that call is in a different FFI library (the
matching one, not the mock server one) so I can't call it. It'll need
exposing on the mock server FFI so the C# side can call it I think

mefellows:
They're re-exported there already: https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_mock_server_ffi/src/lib.rs#L73

adamrodger:
Ah is that not in the latest release? I checked out the source at the tag
of the latest release and didn't see it

adamrodger:
Found it 👍  And now the mock server logs work 🎉 

image

adamrodger:
Oh no - the tests all pass locally for me on Windows, but in the CI they fail on Linux because the `log_to_buffer` function doesn't exist on the `pact_mock_server_ffi` library:
https://github.com/pact-foundation/pact-net/pull/310/checks?check_run_id=3022203564#step:8:101

But they definitely pass locally on Windows:
image

@mefellows Is it possible that the Linux library somehow hasn't had this function re-exported? Or that the `pact-reference` CI has somehow uploaded the wrong FFI library to the release for Linux?

adamrodger:
Yeah, it really looks to me like the Linux library doesn't re-export that function. I really hope that's not a problem with the CI somehow uploading an old version or something.

On Linux those func simply don't exist on the binary:

image

Where on Windows they do:

image

adamrodger:
Raised https://github.com/pact-foundation/pact-reference/issues/122 about the Linux FFI library missing those functions

yanncourtel:
I don't see anything else to say on the PR.

Where are we on the log FFI? Is it working with the new version of the FFI? If so, I'd like to verify everything works on the messaging branch. Thanks a lot.

mefellows:
Thanks Yann. The single FFI is now available at https://github.com/pact-foundation/pact-reference/releases/tag/libpact_ffi-v0.0.0, which fixes the bug. You will need to move to this for this PR to work, I'd say, given the details of https://github.com/pact-foundation/pact-reference/issues/122.

adamrodger:
I'll be raising another PR shortly to upgrade to the new combined FFI library, and then I'll amend this PR to use the new logging functionality from there.

adamrodger:
Superseded by #313

PR #312, opened 2021-08-10
feat!: Replace separate FFI libs with new combined FFI lib

adamrodger:
I've moved the consumer and verifier interop definitions into a single file to make it obvious that the library itself is now a single thing, but maintained the separate interfaces/classes to encapsulate the logic required in each (such as handing the int error codes that you get back from some calls).

### DIFF
--- a/src/PactNet.Native/Interop/NativeInterop.cs
+++ b/src/PactNet.Native/Interop/NativeInterop.cs
@@ -1,0 +1,76 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace PactNet.Native.Interop
+{
+    /// <summary>
+    /// Interop definitions to the Pact FFI library
+    /// </summary>
+    internal static class NativeInterop
+    {
+        private const string dllName = "pact_ffi";
+
+        /// <summary>
+        /// Static initialiser for the Pact FFI library
+        /// </summary>
+        static NativeInterop()
+        {
+            // TODO: Make this configurable and specified by the user
+            LogToBuffer(LevelFilter.Debug);
+        }
+
+        [DllImport(dllName, EntryPoint = "pactffi_log_to_buffer")]
+        public static extern int LogToBuffer(LevelFilter levelFilter);
+
+        [DllImport(dllName, EntryPoint = "pactffi_create_mock_server_for_pact")]
+        public static extern int CreateMockServerForPact(PactHandle pact, string addrStr, bool tls);
+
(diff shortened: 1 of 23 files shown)


src/PactNet.Native/Interop/NativeInterop.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace PactNet.Native.Interop
+{
+    /// <summary>
+    /// Interop definitions to the Pact FFI library
+    /// </summary>
+    internal static class NativeInterop
```
yanncourtel, line 9: Remember when I told you we need to split the interop into a separate file XD good one!

src/PactNet.Native/NativePactVerifier.cs
```diff
@@ -8,23 +7,13 @@ namespace PactNet.Native
     /// </summary>
     internal class NativePactVerifier : IVerifierProvider
     {
-        /// <summary>
-        /// Static constructor for <see cref="NativePactVerifier"/>
-        /// </summary>
-        static NativePactVerifier()
```
yanncourtel, line 14: Why removing this? Is it because of the way we will handle the logs going forward?

src/PactNet.Native/NativeMockServer.cs
```diff
@@ -8,19 +9,9 @@ namespace PactNet.Native
...
     internal class NativeMockServer : IMockServer
     {
-        /// <summary>
-        /// Static constructor for <see cref="NativeMockServer"/>
-        /// </summary>
-        static NativeMockServer()
-        {
-            // TODO: make this configurable somehow, except it applies once for the entire native lifetime, so dunno
-            Environment.SetEnvironmentVariable("LOG_LEVEL", "DEBUG");
-            Interop.Init("LOG_LEVEL");
```
yanncourtel, line 18: Are the logs handled somewhere else or it's a proactive change?

src/PactNet.Native/NativePactVerifier.cs
adamrodger: It's moved into the combined interop file, then in future we can allow the user to specify the log level. I've

src/PactNet.Native/NativeMockServer.cs
adamrodger: The logs part is in the other PR so that this is a single change just to swap to the new FFI. I'll refactor the other one into this once it's merged, because the other doesn't work on Linux without this change.

PR #313, opened 2021-08-12
feat: Mock server logs in consumer test output

adamrodger:
image


yanncourtel (approved):
Have you been able to see the actual logs in the console? I'm saying that because from the latest version (non unified FFI) I wasn't able to.

adamrodger:
@yanncourtel yeah I added a screenshot above showing the log messages. They're a bit "noisy" but as good as we're gonna get 😂

PR #315, opened 2021-08-23
feat: Add V3 messaging consumer support

yanncourtel:

NOTE : this work is following this pull request https://github.com/pact-foundation/pact-net/pull/309 and is only the consumer side.

On the consumer side:
-> integration of the new interop methods for messaging.
-> add new builder for creating messaging interactions.
-> add extensions methods to make it easy to initiate a message interaction.
-> add simple build method to create a messaging pact file.
-> add unit tests


samples/EventImporter/Consumer.Tests/EventImporterTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            };
+
+            IMessagePactV3 v3 = MessagePact.V3("Event API Consumer V3 Message", "Event API V3 Message", _config);
+            _messagePact = v3.UsingNativeBackend();
+        }
+
+        [Fact]
+        public void GetAllEvents_FromQueue_Should_CreatePact_WithMessages()
+        {
+            var expected = new List<dynamic>
```
adamrodger, line 46: Would we genuinely expect people to define the event body outside of the interaction definition? Seems a bit odd to do it that way - at first I didn't even realise what this was being used for.

samples/EventImporter/Consumer.Tests/pacts/Event API Consumer V3 Message-Event API V3 Message.json
```diff
@@ -0,0 +1,106 @@
...
+            "combine": "AND",
+            "matchers": [
+              {
+                "match": "type"
+              }
+            ]
+          }
+        }
+      },
+      "metadata": {},
```
adamrodger, line 42: The test specified some metadata but this appears to be empty.

samples/EventImporter/Consumer/EventsWorker.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            var operationSuccessful = events.Aggregate(true, (current, @event) => current & _eventHandler.ProcessEvent(@event));
+
+            if (!operationSuccessful)
+            {
+                throw new ProcessEventException();
+            }
+        }
+
+        public void DispatchEvent(Event eventSingle)
```
adamrodger, line 32: This example is a bit confusing - why would it have separate methods for processing single or multiple events? In a real application, wouldn't you just call the List version but with only a single element inside?

src/PactNet.Native/NativeMessagePactBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+        private readonly MessagePactHandle pact;
+        private readonly PactConfig config;
+        private MessageHandle message;
+
+        /// <summary>
+        /// Initialises a new instance of the <see cref="NativeMessagePactBuilder"/> class.
+        /// </summary>
+        /// <param name="server">Mock server</param>
+        /// <param name="pact">the messagePact handle</param>
+        /// <param name="config">the messagePact configuration</param>
```
adamrodger, line 25: I think some find and replace has mangled these docs a bit

src/PactNet.Native/NativeMessagePactBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+        #region Internal Methods
+
+        /// <summary>
+        /// Add a new message to the messagePact
+        /// </summary>
+        /// <param name="description">Message description</param>
+        /// <returns>Fluent builder</returns>
+        internal IMessageBuilderV3 ExpectsToReceive(string description)
+        {
+            this.message = this.server.NewMessage(pact, "default message");
```
adamrodger, line 54: I think that needs to be the name of the provider, not a hard-coded string

src/PactNet.Native/NativeMessagePactBuilder.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        /// <summary>
+        /// Verify a message is read and handled correctly and write the messagePact
+        /// </summary>
+        /// <param name="handler">The method using the message</param>
+        public void Verify<T>(Action<T> handler)
+        {
+            try
+            {
+                var content = JsonConvert.DeserializeObject<NativeMessage>(this.server.Reify(message));
```
adamrodger, line 83: I'm a bit confused by this bit - why do message pacts take metadata, content types, headers, etc. but then we only ever use the message body? What uses those extra fields?

src/PactNet.Native/NativeMockServer.cs
```diff
@@ -103,5 +89,60 @@ public bool WithRequestBody(InteractionHandle interaction, string contentType, s
...
+        public string Reify(MessageHandle message)
+        {
+            var allocatedString = Marshal.PtrToStringAnsi(NativeInterop.MessageReify(message));
+            return allocatedString;
+        }
+
+
+        #endregion Message Interaction Model Support
+
+        private void VerifyResult(Func<int> func)
```
adamrodger, line 132: Are we certain that the request/response and message pact styles will always have the same response types/codes?

src/PactNet.Native/NativeMockServer.cs
```diff
@@ -103,5 +89,60 @@ public bool WithRequestBody(InteractionHandle interaction, string contentType, s
...
+
+            if (result != 0)
+            {
+                throw result switch
+                {
+                    1 => new InvalidOperationException("The pact reference library panicked"),
+                    2 => new InvalidOperationException("The pact file could not be written"),
+                    3 => new InvalidOperationException("A mock server with the provided port was not found"),
+                    _ => new InvalidOperationException($"Unknown error from backend: {result}")
+                    };
```
adamrodger, line 144: Formatting has gone wonky here

src/PactNet.Native/PactExtensions.cs
```diff
@@ -9,7 +9,7 @@ namespace PactNet.Native
     public static class PactExtensions
     {
         /// <summary>
-        /// Establish a new pact using the native backend
+        /// Establish a new messagePact using the native backend
```
adamrodger, line 12: Find and replace appears to have mangled this file also

src/PactNet.Native/PactExtensions.cs
```diff
@@ -51,17 +51,51 @@ public static IPactBuilderV3 UsingNativeBackend(this IPactV3 pact, int? port = n
         }
 
         /// <summary>
-        /// Initialise a new pact on the server with the correct version
+        /// Establish a new messagePact using the native backend
+        /// </summary>
+        /// <param name="messagePact">Pact details</param>
+        /// <returns>Pact builder</returns>
+        /// <remarks>
+        /// If multiple mock servers are started at the same time, you must make sure you don't supply the same port twice.
```
adamrodger, line 59: I don't think these docs make sense for message pacts do they? I don't think it starts a mock server at all.

src/PactNet.Native/PactExtensions.cs
```diff
@@ -51,17 +51,51 @@ public static IPactBuilderV3 UsingNativeBackend(this IPactV3 pact, int? port = n
...
+        /// <param name="messagePact">Pact details</param>
+        /// <param name="server">Server</param>
+        /// <param name="version">Spec version</param>
+        /// <returns>Initialised message messagePact handle</returns>
+        private static MessagePactHandle InitialiseMessage(NativeMockServer server, IMessagePact messagePact, PactSpecification version)
+        {
+            PactHandle handle = server.NewPact(messagePact.Consumer, messagePact.Provider);
+            server.WithSpecification(handle, version);
+
+            return server.NewMessagePact(messagePact.Consumer, messagePact.Provider);
```
adamrodger, line 98: Hang on - it creates a request/response pact and then immediately creates a new message pact? I'm not sure that's what you meant to do. Also the request builder class above also called `NewMessagePact` so it'll effectively overwrite this anyway.

src/PactNet/IMessagePactBuilder.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using System.Threading.Tasks;
+
+namespace PactNet
+{
+    public interface IMessagePactBuilder
```
adamrodger, line 6: Missing docs

src/PactNet/IMessagePactBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Add metadata information to message pact
+        /// </summary>
+        /// <param name="namespace">the parent configuration section</param>
+        /// <param name="name">the metadata field value</param>
+        /// <param name="value">the metadata field value</param>
+        /// <returns>Fluent builder</returns>
+        IMessagePactBuilderV3 WithPactMetadata(string @namespace, string name, string value);
+        //INTEROP =>
```
adamrodger, line 41: Left over commented out code that I suspect needs removing

src/PactNet/IMessagePactBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Verify a message is read and handled correctly
+        /// </summary>
+        /// <param name="handler">The method using the message</param>
+        Task VerifyAsync<T>(Func<T, Task> handler);
+    }
+
+    /// <summary>
+    /// Build up a mock message for a v3 messagePact
```
adamrodger, line 22: More `messagePact` find and replace problems I think.

src/PactNet/Pact.cs
```diff
@@ -53,13 +53,13 @@ private Pact(string consumer, string provider, PactConfig config)
                 throw new ArgumentException("Please provide a valid provider name", nameof(provider));
             }
 
-            this.Consumer = consumer;
```
adamrodger, line 56: I think this file just needs reverting.

adamrodger (changes requested):
I've not reviewed the tests in detail given there are probably changes requires to the main code body beforehand.

src/PactNet.Native/NativeMessagePactBuilder.cs
```diff
@@ -45,21 +45,21 @@ IMessagePactBuilderV3 IMessagePactBuilderV3.WithPactMetadata(string @namespace,
...
         /// <summary>
-        /// Add a new message to the messagePact
+        /// Add a new message to the message pact
         /// </summary>
         /// <param name="description">Message description</param>
         /// <returns>Fluent builder</returns>
         internal IMessageBuilderV3 ExpectsToReceive(string description)
         {
-            this.message = this.server.NewMessage(pact, "default message");
+            this.message = this.server.NewMessage(pact, description);
```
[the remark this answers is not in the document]
adamrodger: Yep I've noticed my mistake now 👍 I thought MessagHandle and MessagePactHandle were the same thing because they have similar names

adamrodger (commented):
There are some compiler warnings on the async tests because the VerifyAsync has no await. I think if we add those then we're ready to merge 👍

yanncourtel:
> There are some compiler warnings on the async tests because the VerifyAsync has no await. I think if we add those then we're ready to merge 👍

Fair enough. Thanks for reviewing on a Saturday :)
I added a simple Task.Run to do the trick as we are not checking the handler itself but that the verifier async methods works. I'll add also the similar unit test for non messaging verify async on another PR.

tests/PactNet.Native.Tests/NativePactMessageBuilderTests.cs
```diff
@@ -123,7 +124,7 @@ public void VerifyAsync_Should_Write_Pact_File()
             SetServerReifyMessage(JsonConvert.SerializeObject(content.ToNativeMessage()));
 
             //Act
-            this.messagePact.VerifyAsync<MessageModel>(async _ => { });
+            this.messagePact.VerifyAsync<MessageModel>(async _ => await Task.Run(() => true));
```
adamrodger, line 127: These would read better as non-async lambdas returning Task.FromResult(true)

I think the test itself should also be async Task instead of void, and then this VerifyAsync would itself be awaited. I'm surprised there aren't more compiler warnings here? Usually if something returns Task but you don't await it then you'll get a warning. In real usage we'd expect callers to be using await here for sure.

yanncourtel: Yeah, looks like I induced a deadlock by not awaiting anything in the handler. 

"I'm surprised there aren't more compiler warnings here?"

Actually no because the Action or Func can either run sync or async so the compiler is fine with that, but deadlocks at runtime... I had a pair or headaches because of that! 

I will test the verify async in the integration tests (the sample projet) the same way we do for the http interaction model. That will be closer to a real usage.

yanncourtel: Seems to work fine now

adamrodger (approved):

PR #316, opened 2021-08-24
feat: Use matchers with request/response headers

edanielsen:
An attempt at supporting matchers for request & response headers (as it is supported by the Rust engine).

I'm sure there is some missing test coverage somewhere, just not sure where.

I implemented this because I selfishly need it...


src/PactNet.Native/NativeRequestBuilder.cs
```diff
@@ -78,8 +78,8 @@ IRequestBuilderV2 IRequestBuilderV2.WithQuery(string key, string value)
         /// <param name="key">Header key</param>
         /// <param name="value">Header value</param>
         /// <returns>Fluent builder</returns>
-        IRequestBuilderV2 IRequestBuilderV2.WithHeader(string key, string value)
-            => this.WithHeader(key, value);
+        IRequestBuilderV2 IRequestBuilderV2.WithHeader(string key, dynamic value)
```
adamrodger, line 81: Ah, we'd specifically moved away from `dynamic` because it's very difficult to get that to play nicely with the Rust code. I'd prefer an overload which took an `IMatcher` perhaps? The API is designed specifically for principle of least surprise, and `dynamic` lets you do some crazy stuff where in reality we'd only support `string` and `IMatcher`.

tests/PactNet.Native.Tests/NativeRequestBuilderTests.cs
```diff
@@ -93,6 +93,14 @@ public void WithQuery_RepeatedQuery_SetsIndex()
             this.mockServer.Verify(s => s.WithQueryParameter(this.handle, "other", "value", 0));
         }
 
+        [Fact]
+        public void WithHeader_WhenCalledWithMatcher_AddsSerialisedHeaderParam()
```
adamrodger, line 97: If we rename the existing test to `WithHeader_String_...` then this can be `WithHeader_Matcher_...`

tests/PactNet.Native.Tests/NativeRequestBuilderTests.cs
```diff
@@ -93,6 +93,14 @@ public void WithQuery_RepeatedQuery_SetsIndex()
...
+        {
+            this.builder.WithHeader("name", PactNet.Matchers.Match.Regex("header", "^header$"));
+
+            this.mockServer.Verify(s => s.WithRequestHeader(this.handle, "name", "{\"pact:matcher:type\":\"regex\",\"value\":\"header\",\"regex\":\"^header$\"}", 0));
```
adamrodger, line 101: I think the JSON probably needs splitting out into an `expected` var so it's easier to read given the long line it creates.

tests/PactNet.Native.Tests/NativeRequestBuilderTests.cs
```diff
@@ -93,6 +93,14 @@ public void WithQuery_RepeatedQuery_SetsIndex()
...
+        }
+
```
adamrodger, line 103: Needs a test which makes sure it respects the serialisation settings, although given we only support matchers is that necessary? I'm not sure you could customise the serialisation settings for those and still have them compatible with the Rust core.

adamrodger (changes requested):

src/PactNet.Native/NativeRequestBuilder.cs
edanielsen: Fair point, I'm definitely not a .Net developer :)

tests/PactNet.Native.Tests/NativeRequestBuilderTests.cs
edanielsen: 👍

tests/PactNet.Native.Tests/NativeRequestBuilderTests.cs
edanielsen: I remove the option to customise this

src/PactNet.Native/NativeRequestBuilder.cs
```diff
@@ -163,6 +173,15 @@ IRequestBuilderV3 IRequestBuilderV3.WithQuery(string key, string value)
...
             => this.WithHeader(key, value);
 
+        /// <summary>
+        /// Add a request header
+        /// </summary>
+        /// <param name="key">Header key</param>
+        /// <param name="valueMatcher">Header value matcher</param>
+        /// <returns>Fluent builder</returns>
+        IRequestBuilderV3 IRequestBuilderV3.WithHeader(string key, IMatcher valueMatcher)
+            => this.WithHeader(key, valueMatcher);
```
adamrodger, line 183: I'm gonna disagree with @yanncourtel a bit here I think. I think that var name is quite clunky. I would've just called it matcher.

tests/PactNet.Native.Tests/NativeRequestBuilderTests.cs
```diff
@@ -94,15 +94,41 @@ public void WithQuery_RepeatedQuery_SetsIndex()
...
+        public void WithHeader_String_WhenCalled_AddsHeaderParam()
         {
             this.builder.WithHeader("name", "value");
 
             this.mockServer.Verify(s => s.WithRequestHeader(this.handle, "name", "value", 0));
         }
 
         [Fact]
-        public void WithHeader_RepeatedHeader_SetsIndex()
+        public void WithHeader_String_RepeatedHeader_SetsIndex()
```
adamrodger, line 131: The naming convention for the tests is Method_PreCondition_PostCondition so I'd call this WithHeader_RepeatedStringHeader_SetsIndex.

We'll need additional tests for RepeatedMatcherHeader as well, and maybe one that does both a string and a matcher together.

Edit: sorry just seen the matcher tests below, so just the naming thing

adamrodger (changes requested):

src/PactNet.Native/NativeRequestBuilder.cs
yanncourtel: My point was consistency across the same type of methods. One of the variable was called valueBuilder which I found it odd (prob copy/paste thingy). You are right, matcher is actually a better name.

src/PactNet.Remote/RemoteRequestBuilder.cs
```diff
@@ -60,7 +60,7 @@ IRequestBuilderV2 IRequestBuilderV2.WithQuery(string key, string value)
         /// <param name="key">Header key</param>
         /// <param name="value">Header value</param>
         /// <returns>Fluent builder</returns>
-        IRequestBuilderV2 IRequestBuilderV2.WithHeader(string key, string value)
+        IRequestBuilderV2 IRequestBuilderV2.WithHeader(string key, dynamic value)
```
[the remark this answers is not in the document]
edanielsen: Yes, this is refactoring gone wrong

PR #318, opened 2021-08-30
feat: Updated README for v4.0.0

ajpatterson7:


README.md
```diff
@@ -1,492 +1,271 @@
...
-            .PactUri("http://pact-broker/pacts/provider/Something%20Api/consumer/Consumer/latest", pactUriOptions) //With options decribed above
-            //or (if you're using the Pact Broker, you can use the various different features, including pending pacts)
-            .PactBroker("http://pact-broker", uriOptions: pactUriOptions, enablePending: true, consumerVersionTags: new   List<string> { "master" }, providerVersionTags: new List<string> { "master" }, consumerVersionSelectors: new List<VersionTagSelector> { new VersionTagSelector("master") })
-            .Verify();
+        ServerUri = new Uri("http://localhost:9222");
+        server = Host.CreateDefaultBuilder()
+                    .ConfigureWebHostDefaults(webBuilder =>
+                    {
+                        webBuilder.UseUrls(ServerUri.ToString());
+                        webBuilder.UseStartup<TestStartup>();
```
adamrodger, line 179: I'd put a comment against this section explaining that you are expected to start your own web host that serves your real API., and this is a simple example of how to do that.

People frequently seem to ask for in-memory `TestHost` support (from the `Microsoft.AspNetCore.TestHost` package) and we simply can't support that.. Something like:

------

In order for Pact to call your API with the expectations from the contract, your API must be running on a real locally-accessible port using a real `Host`. This simple example shows how you can do that.

We are frequently asked whether a `TestHost` could be used instead (from the `Microsoft.AspNetCore.TestHost` package), but this isn't possible because the core Pact library is not running in .Net and therefore can't use an in-memory `HttpClient`.

README.md
```diff
@@ -1,492 +1,271 @@
...
 
-    public void WriteLine(string line)
-    {
-        _output.WriteLine(line);
+        // Act / Assert
+        IPactVerifier pactVerifier = new PactVerifier(config);
+        pactVerifier
+            .ServiceProvider("Something API", fixture.ServerUri)
+            .HonoursPactWith("Something API Consumer")
+            .FromPactFile(new FileInfo(pactPath))
```
adamrodger, line 231: The original docs showed how you've got a few different options for this, such as using Pact Broker to retrieve your pacts instead of from a local file. I think it would be useful to keep that because we don't document those sorts of things anywhere else.

README.md
```diff
@@ -1,492 +1,271 @@
...
+            .UponReceiving("A GET request to retrieve the something")
+                .Given("There is a something with id 'tester'")
+                .WithRequest(HttpMethod.Get, "/somethings/tester")
+                .WithHeader("Accept", "application/json")
+            .WillRespond()
+                .WithStatus(HttpStatusCode.OK)
+                .WithHeader("Content-Type", "application/json; charset=utf-8")
+                .WithJsonBody(new
+                {
+                    // NOTE: These properties are case sensitive!
```
adamrodger, line 139: This isn't necessarily true - the JSON settings that you supply within the original `PactConfig` determine how your data will be serialised, with one very popular option being to control the casing of properties.

Typically I'd expect people to use .Net style UpperCamelCase for the property names in the example objects, but set the JSON settings to serialise to lowerCamelCase to be more JSON compliant.

Perhaps it's worth having an extra section which shows how you can customise the JSON settings and changing this comment to make it clear that's what determines what these objects look like?

README.md
```diff
@@ -1,492 +1,271 @@
...
+        // Arrange
+        PactBuilder
+            .UponReceiving("A GET request to retrieve the something")
+                .Given("There is a something with id 'tester'")
+                .WithRequest(HttpMethod.Get, "/somethings/tester")
+                .WithHeader("Accept", "application/json")
+            .WillRespond()
+                .WithStatus(HttpStatusCode.OK)
+                .WithHeader("Content-Type", "application/json; charset=utf-8")
+                .WithJsonBody(new
```
adamrodger, line 137: Typically we'd expect people to be using `Matcher` instead of a fixed object here wouldn't we? We don't document those anywhere else currently so it would be good to have the main example using a `Matcher` and explaining why that makes sense.

samples/ReadMe/Provider.Tests/SomethingApiFixture.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        public void Dispose()
+        {
+            server.Dispose();
+
+            // Clean out leftover data
+            foreach (var dataFile in Directory.GetFiles(Path.Combine("..", "..", "..", "data")))
+            {
+                File.Delete(dataFile);
```
adamrodger, line 33: This is probably going to make the tests annoying to run because you'll always need to remember to run the consumer test first and then the provider test later to delete the file (which makes CI a bit awkward given they're sharing state effectively).

The other samples check the pact file in and never delete it.

samples/ReadMe/Provider.Tests/SomethingApiTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            };
+
+            string pactPath = Path.Combine("..",
+                                           "..",
+                                           "..",
+                                           "..",
+                                           "pacts",
+                                           "Something API Consumer-Something API.json");
+
+            // Thread.Sleep(TimeSpan.FromMinutes(1));
```
adamrodger, line 46: Remove commented out line

samples/ReadMe/Provider.Tests/TestStartup.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            inner.ConfigureServices(services);
+        }
+
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            app.UseMiddleware<ProviderStateMiddleware>()
+                .UseMiddleware<AuthorizationTokenMiddleware>();
```
adamrodger, line 25: I wonder if it might be slightly overkill adding auth to this simple example. It adds a fair bit of semi-complex code which doesn't really relate to the actual core example of how to write Pact tests, whereas the other samples cover auth as they're more broadly scoped.

samples/ReadMe/Provider/Startup.cs
```diff
@@ -0,0 +1,67 @@
...
+                });
+
+            services.AddAuthentication(options =>
+                {
+                    options.DefaultAuthenticateScheme = JwtBearerDefaults.AuthenticationScheme;
+                    options.DefaultChallengeScheme = JwtBearerDefaults.AuthenticationScheme;
+                })
+                .AddJwtBearer(options =>
+                {
+                    options.Authority = "https://something.api.com";
```
adamrodger, line 44: Canonically you're supposed to use the `example.org` domain for made up URLs just in case you accidentally end up sending traffic to a real domain.

samples/ReadMe/Provider/Readme.Provider.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="5.0.6" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Swashbuckle.AspNetCore" Version="5.6.3" />
```
adamrodger, line 10: The sample csproj files all reference `Swashbuckle.AspNetCore` but none actually add Swagger to the API. We should remove the unused reference.

samples/ReadMe/pacts/Something API Consumer-Something API.json
```diff
@@ -0,0 +1,44 @@
+{
+  "consumer": {
+    "name": "Something API Consumer"
+  },
+  "interactions": [
+    {
+      "description": "A GET request to retrieve the something",
+      "providerStates": [
+        {
+          "name": "There is a something with id 'tester'"
```
adamrodger, line 10: We should probably use the "provider state with param" option instead of hard-coding the ID in here, since we'd expect that to be best practice from v3 onwards.

adamrodger (changes requested):

README.md
```diff
@@ -1,492 +1,271 @@
...
+![logo](https://user-images.githubusercontent.com/53900/121775784-0191d200-cbcd-11eb-83dd-adc001b94519.png)
 
-PactNet primarily provides a fluent .NET DSL for describing HTTP requests that will be made to a service provider and the HTTP responses the consumer expects back to function correctly.  
-In documenting the consumer interactions, we can replay them on the provider and ensure the provider responds as expected. This basically gives us complete test symmetry and removes the basic need for integrated tests.  
-PactNet also has the ability to support other mock providers should we see fit.
+# Pact Net
 
-PactNet is Version 2.0 compliant, and we now use the [Ruby standalone engine](https://github.com/pact-foundation/pact-ruby-standalone) as we move towards a common core approach. To enable Version 2.0 support,  make sure you supply a `PactConfig` object with `SpecificationVersion = "2.0.0"` when creating the `PactBuilder`.  
+[![Build status](https://ci.appveyor.com/api/projects/status/5h4t9oerlhqcnwm8/branch/master?svg=true)](https://ci.appveyor.com/project/PactFoundation/pact-net/branch/master)
```
adamrodger, line 7: Oh forgot to mention - the CI is now done in GitHub Actions instead of AppVeyor. We should probably update the badge. Ci is here: https://github.com/pact-foundation/pact-net/actions/workflows/ci.yml

PactNet.sln
```diff
@@ -1,202 +1,264 @@
-﻿
```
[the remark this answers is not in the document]
ajpatterson7: Woah, no idea what happened with this diff. I'll take a look and try to clean that up. Every line?? D:

README.md
```diff
@@ -1,492 +1,271 @@
...
-public class ProviderState
-{
-    public string Consumer { get; set; }
-    public string State { get; set; }
-}
-```
+                // NOTE: We default to using a ConsoleOutput,
+                // however xUnit 2 does not capture the console
+                // output, so a custom outputter is required.
+                new XUnitOutput(output),
```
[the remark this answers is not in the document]
ajpatterson7: I'll look to clear this up, as we include a definition for it in the sample project this is pulled from. It's good to call out that this is not a part of Pact.Net

samples/ReadMe/Provider.Tests/TestStartup.cs
ajpatterson7: Let's see if I can pull out all the auth code and still make it work. Agreed that it adds a lot of complexity and no value to the example.

README.md
```diff
@@ -1,492 +1,271 @@
...
-**A few others things to note:**
-1. When using `Match.Regex` you must supply a valid Ruby regular expression, as we currently use the Ruby core engine.
+## Documentation
 
-### Installing
+This readme offers an basic introduction to the library. The full documentation for Pact Net and the rest of the framework is available at https://docs.pact.io/.
 
-Via Nuget  
+-   [Installation](#installation)
+-   [Consumer Testing](./docs/consumer.md)
```
mefellows, line 54: These links currently go to nowhere - did you forget to check them in?

README.md
```diff
@@ -1,492 +1,271 @@
...
 
-**Windows**  
-https://www.nuget.org/packages/PactNet.Windows  
-`Install-Package PactNet.Windows`
+### Tutorial (60 minutes)
 
-**OSX**  
-https://www.nuget.org/packages/PactNet.OSX  
-`Install-Package PactNet.OSX`
+Learn everything in Pact Net in 60 minutes: // TODO
```
mefellows, line 63: Can link to here: https://github.com/DiUS/pact-workshop-dotnet-core-v3/

README.md
```diff
@@ -1,492 +1,271 @@
...
-#### 1. Build your client
-Which may look something like this.
+Via Nuget
 
-```c#
-public class SomethingApiClient
-{
-  private readonly HttpClient _client;
+**Windows**
+https://www.nuget.org/packages/PactNet.Windows
```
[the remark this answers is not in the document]
mefellows: Link to https://github.com/DiUS/pact-workshop-dotnet-core-v3/#add-the-pacflow-nuget-repository-to-visual-studionugetconfig for now?

README.md
mefellows: I'd recommend making this example point to a Pact Broker with a good default configuration, because people will almost certainly copy paste.

I say that knowing full well the Golang example doesn't do this, and I'm going to fix that.

Doesn't need to be in this PR though.

samples/ReadMe/Provider/Startup.cs
ajpatterson7: Good to know, I'll update. Thanks!

samples/ReadMe/pacts/Something API Consumer-Something API.json
ajpatterson7: Could you provide a sample here? This looks similar to the pact-go docs where they used "A user with ID 10 exists".

README.md
ajpatterson7: Missed this section during my edits. I'll remove for now and can treat this as a to-do list of later docs

PactNet.sln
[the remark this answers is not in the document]
ajpatterson7: Line endings! Had the wrong git setting

ajpatterson7:
All comments addressed or followed up on - left conversations unresolved to give reviewers opportunity to verify changes meet expectation. Please let me know if it's preferred for author to mark as resolved :)

PactNet.sln
```diff
@@ -1,264 +1,232 @@
...
```
adamrodger: The file still seems to have line ending problems. I think `.sln` files have to have `\r\n` line endings if I remember correctly, so if you've made this commit on a non-Windows machine that may be the problem.

Perhaps add a `.gitattributes` file to force `.sln` to have `CRLF` line endings?

adamrodger (commented):
Just need the solution file sorting out or it may not work on Windows (?). It's hard to tell from this diff what's gone wrong there though. Perhaps a `.gitattributes` will sort it.

PactNet.sln
ajpatterson7: I'm not seeing any line ending issues at this point, there were a number of errors that had to fixed with a rebase to get the EventImporter sample that are now addressed. The diff is showing just the expected added lines for me, let me know if you still see the line endings showing up wrong.

adamrodger (approved):

Issue #323, opened 2021-09-19
Poor Consumer Test Performance on Windows


adamrodger:
Consumer tests with the new Rust FFI have very poor performance when calling the mock server via `HttpClient` on Windows. This is quite a large regression on the Ruby backend performance, and one of the key benefits of the Rust FFI backend is supposed to be increased performance.

## Demonstration of performance

It typically takes around 2 seconds for the `HttpResponse` to come back, which means every consumer test takes a minimum of 2 seconds. It's not uncommon to have dozens of consumer tests per solution, and testing libraries typically won't run tests within the same file/collection in parallel, and so it would result in the consumer test run taking minutes to complete.

Here's an example:

![image](https://user-images.githubusercontent.com/508586/133938806-591a2359-3362-442a-ae2a-ee08d4c20c39.png)

And with some logging added to show where the problem happen:

![image](https://user-images.githubusercontent.com/508586/133938993-d8a1f902-055c-470f-bd22-a1dc29c74415.png)

![image](https://user-images.githubusercontent.com/508586/133938983-410d4704-dd1a-45dd-82e0-5e2ad682c6a6.png)

## Only on Windows

It's worth noting that this only seems to happen on Windows. On Linux the consumer tests run extremely quickly. Check [recent CI runs](https://github.com/pact-foundation/pact-net/actions) to confirm, but typically the Ubuntu run finishes in a fraction of the time of the Windows run. I have confirmed locally that the tests are extremely slow on my Windows machine as well, so it's not some issue with GitHub Actions.

## Replace the Pact Mock Server with an echo server

To confirm that it's not caused by `HttpClient` itself, I ran a local [Mock Server](https://www.mock-server.com/), like this:

```bash
docker run --rm -it -p 1080:1080 mockserver/mockserver
```

I set up the following hard-coded expectation (which is one of the responses expected by one of the example consumer tests) and hard-coded the same example consumer test to use the echo server on http://localhost:1080 instead of the Pact Mock Server:

```bash
curl -v -X PUT "http://localhost:1080/mockserver/expectation" -d '{
    "httpRequest": {
        "path": "/events"
    },
    "httpResponse": {
        "statusCode": 200,
        "headers": {
            "content-type": ["application/json"]
        },
        "body": [{
            "eventId": "45d80d13-d5a2-48d7-8353-cbb4c0eaabf5",
            "eventType": "SearchView",
            "timestamp": "2014-06-30T01:37:41.0660548"
        }, {
[...]

PR #324, opened 2021-09-19
fix: Massively improve consumer test performance on Windows (#323)

adamrodger:
Just a 14.3s improvement from such a small typo 😱 

Before:
image

After:
image


adamrodger:
Note: I've also fixed the build issue on Linux - the solution file was referencing the new samples with the wrong case in the path (`Samples` instead of `samples`). I didn't see much point making that a separate pull request.

mefellows:
Love it 👏 .

So glad you've found the source of this - nicely done!

mefellows (approved):

src/PactNet.Native/NativePactBuilder.cs
```diff
@@ -141,7 +141,8 @@ private Uri StartMockServer()
             // TODO: add TLS support
             int serverPort = this.server.CreateMockServerForPact(this.pact, address, false);
 
-            var uri = new Uri($"http://{this.host}:{serverPort}");
+            var mockServerUrl = $"http://{hostIp}:{serverPort}";
+            var uri = new Uri(mockServerUrl);
```
mefellows, line 145: Minor, but any reason to not just inline the return i.e. `return new Uri(...)`?

yanncourtel:
> Note: I've also fixed the build issue on Linux - the solution file was referencing the new samples with the wrong case in the path (`Samples` instead of `samples`). I didn't see much point making that a separate pull request.

Wow! The simple referencing caused such a performace issue! Damn, that's amazing

yanncourtel (approved):

src/PactNet.Native/NativePactBuilder.cs
adamrodger: Because I never would've tracked down the problem otherwise 😆 Having those are separate vars allows you to eyeball them with the debugger, but no perf hit or anything in a Release build.

PR #326, opened 2021-09-27
Feature/messaging provider

yanncourtel:

This part is the provider side

On the provider side:
-> add middleware to act as a messaging proxy server (so far only netstandard 2.0 compatible)
-> add pact message verifying support
-> add scenario builder to emulate the push of a message on the provider side (see sample app: EventImporterProviderTests).

Regarding the tests:
-> add unit tests.
-> add some integration tests.
-> add new sample application for integrated testing with a messaging pact.


src/PactNet.Native/MessageScenarioBuilder.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class MessageScenarioBuilder : IMessageScenarioBuilder
+    {
+        private MessageScenarioBuilder()
+        {
+        }
+
+        /// <summary>
+        /// Handles to create a new scenario with scenario builder
+        /// </summary>
+        public static IMessageScenarioBuilder AScenario => new MessageScenarioBuilder();
```
adamrodger, line 17: This feels quite clunky - can't the `WhenReceiving` method below be used instead?

samples/EventImporter/Provider.Tests/AuthorizationTokenReplacementMiddleware.cs
```diff
@@ -0,0 +1,29 @@
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Provider.Tests
+{
+    public class AuthorizationTokenReplacementMiddleware
```
adamrodger, line 6: I'd probably remove all the auth stuff from this example. It just adds noise whereas the other example already shows how to do auth.

samples/EventImporter/Provider.Tests/EventImporterProviderTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                                           "..",
+                                           "..",
+                                           "..",
+                                           "Consumer.Tests",
+                                           "pacts",
+                                           "Event API Consumer V3 Message-Event API V3 Message.json");
+
+            //Setting up scenarios for messaging simulation.
+            //If other provider tests, the unique constraint would need to be handled (the ClearScenarios method of the Scenarios class is available)
+            SetupEventImporterScenarios();
```
adamrodger, line 58: I'm generally confused about this whole thing works, perhaps because it's backwards from the usual request/response type. So the consumer has already defined the pact file, and now the provider is effectively setting up a sort of provider state using the magic strings inside that pact file?

That seems quite brittle, because new interactions created by the consumer immediately need code changes in the provider otherwise it's going to throw exceptions and fail the test.

Perhaps I'm just not understanding how this works, but I've read through it twice now and I don't really get it.

samples/EventImporter/Provider.Tests/TestStartup.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.Configure<MessagingVerifierOptions>(configuration.GetSection("PactMessageMiddleware"));
+
+            this.inner.ConfigureServices(services);
+        }
+
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            app.UseMessaging();
```
adamrodger, line 31: This probably wants to be named something like `UsePactMessaging` so it's not so generic

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <summary>
+    /// Defines the message middleware
+    /// It handles the queries for the verifier when there are message interactions
+    /// </summary>
+    public class MessageMiddleware
+    {
+        /// <summary>
+        /// The default serializer options
+        /// </summary>
+        public static JsonSerializerOptions SerializerOptions => new JsonSerializerOptions
```
adamrodger, line 25: This is using `System.Text.Json` but I'm pretty sure the main `PactNet` repo uses Newtonsoft doesn't it? We should probably make sure we're only using one.

samples/EventImporter/Provider.Tests/TestStartup.cs
```diff
@@ -0,0 +1,36 @@
...
+        public TestStartup(IConfiguration configuration)
+        {
+            this.configuration = configuration;
+
+            this.inner = new Startup(configuration);
+        }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.Configure<MessagingVerifierOptions>(configuration.GetSection("PactMessageMiddleware"));
```
adamrodger, line 24: This should probably be a dedicated extension method like `services.AddPactMessaging(options => { ... })`;

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// </summary>
+        private readonly MessagingVerifierOptions options;
+
+        /// <summary>
+        /// Creates an instance of <see cref="MessageMiddleware"/>
+        /// </summary>
+        /// <param name="options">the middleware options</param>
+        /// <param name="scenarios">the list of scenarios</param>
+        /// <param name="next">the next handle</param>
+        public MessageMiddleware(IOptionsMonitor<MessagingVerifierOptions> options, Scenarios scenarios, RequestDelegate next)
```
adamrodger, line 51: That just needs to take an `IOptions` doesn't it? We're not expecting the options to change one the test is running.

src/PactNet.AspNetCore/PactNet.AspNetCore.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\PactNet.Native\PactNet.Native.csproj" />
+    <ProjectReference Include="..\PactNet\PactNet.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework.StartsWith('netstandard'))">
```
adamrodger, line 12: Everything in here is .Net Standard so these 2 sections should be merged together. I'm not enturely sure we actually need any of these anyway?

src/PactNet.Native/NativeMessagePactBuilder.cs
```diff
@@ -120,5 +114,31 @@ public async Task VerifyAsync<T>(Func<T, Task> handler)
...
+
+        /// <summary>
+        /// Try to read the reified message
+        /// </summary>
+        /// <typeparam name="T">the type of message</typeparam>
+        /// <returns>the message</returns>
+        private T MessageReified<T>()
+        {
+            string reified = this.server.Reify(this.message);
+            NativeMessage content = JsonConvert.DeserializeObject<NativeMessage>(reified);
```
adamrodger, line 128: Yeah this is using Newtonsoft but the other side appears to use System.Text.Json. I imagine that'll cause all kinds of wird incompatibility and unexpected outcomes.

src/PactNet.Native/PactVerifier.cs
```diff
@@ -64,6 +68,36 @@ public IPactVerifier ServiceProvider(string providerName, Uri pactUri)
...
         }
 
+        /// <summary>
+        /// Set the provider details
+        /// </summary>
+        /// <param name="providerName">Name of the provider</param>
+        /// <param name="pactUri">URI of the running service</param>
+        /// <param name="messagingRoute">Route where the messaging interaction is going to be sent</param>
+        /// <returns>Fluent builder</returns>
+        public IPactVerifier MessagingServiceProvider(string providerName, Uri pactUri, string messagingRoute)
```
adamrodger, line 78: This makes the API kind of weird because you could technically call both kids of service provider method, and then I'm not sure what would happen.

src/PactNet.Native/PactVerifier.cs
```diff
@@ -64,6 +68,36 @@ public IPactVerifier ServiceProvider(string providerName, Uri pactUri)
...
+            if (string.IsNullOrWhiteSpace(messagingRoute))
+            {
+                throw new ArgumentException("The messaging route cannot be null or empty");
+            }
+
+            SetProviderHost(providerName, pactUri);
+
+            verifierArgs.Add("--base-path");
+
+            var messageBasePath = pactUri.AbsolutePath != "/" ? pactUri.AbsolutePath + messagingRoute : messagingRoute;
```
adamrodger, line 94: Joining URLs with `+` is probably dangerous, like if the absolute path ends with `/` and then the user specifies `/` at the start of theirs then the URL will end up `http://localhost/foo//bar` which might be break. We should join in a safer way.

src/PactNet.Native/Scenarios.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using PactNet.Models;
+
+namespace PactNet.Native
+{
+    /// <summary>
+    /// Defines the scenarios static list for messaging support
+    /// </summary>
+    public class Scenarios
```
adamrodger, line 11: Is this class supposed to be public? Aren't users supposed to use the builders in the main `PactNet` project to define these?

src/PactNet.Native/Scenarios.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        /// <summary>
+        /// The available scenarios
+        /// </summary>
+        internal readonly List<Scenario> AllScenarios = new List<Scenario>();
```
adamrodger, line 16: Shouldn't this be private?

src/PactNet.Native/Scenarios.cs
```diff
@@ -0,0 +1,104 @@
...
+            AllScenarios.Add(scenario);
+        }
+
+        /// <summary>
+        /// Add multiple scenarios
+        /// </summary>
+        /// <param name="scenarios">the scenario list to add</param>
+        public void AddScenarios(IReadOnlyCollection<Scenario> scenarios)
+        {
+            if (scenarios == null || scenarios.Any() == false)
```
adamrodger, line 43: It's idiomatic to use `!scenarios.Any()` instead of `scenarios.Any() == false`

src/PactNet.Native/Scenarios.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        /// <summary>
+        /// Get a scenario by description 
+        /// </summary>
+        /// <param name="description">the scenario description</param>
+        /// <returns>The scenario</returns>
+        public Scenario GetByDescription(string description)
+        {
+            return AllScenarios.Single(x => x.Description == description);
```
adamrodger, line 93: This will throw a weird LINQ exception where you probably want a more meaningful one, plus add it to the documentation.

src/PactNet.Native/Scenarios.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                throw new ArgumentNullException(nameof(description));
+            }
+
+            if (!Exist(description))
+            {
+                throw new InvalidOperationException($"Scenario \"{description}\" not found. You need to add the scenario first");
+            }
+
+            return AllScenarios.Single(x => x.Description == description).InvokeScenario();
```
adamrodger, line 73: Instead of searching the list twice (both here and in `Exists`) wouldn't it be better to use something like `FirstOrDefault` and then throw on default?

adamrodger (changes requested):

mefellows:
Sorry for the delay @yanncourtel, i'll give this a good look over tomorrow!

src/PactNet.Native/MessageScenarioBuilder.cs
yanncourtel: No that won't do. The only way around would be to put public the constructor and create an instance of the builder. The syntax is just horrible to write.
In the consumer approach, you are using an extensions method to do the heavy lifting for you and since it has different version, it allows more flexibility.

Therefore we can either:
1) use another set of extensions method for scenario setting (maybe by specifying a pact version) 
2) or a static handle, like here, which is a common practice in fluent builders.

samples/EventImporter/Provider.Tests/EventImporterProviderTests.cs
yanncourtel: You've got to think of the whole messaging approach.
The workflow is indeed reversed in the real-life API than it would be for HTTP interaction model.

1) The API producing the resource first pushes it to the queue.
2) The API that subscribed to the queue is reading the queue afterwards asynchronously. 

The problem is in PACT, you cannot reverse the workflow since it's consumer driven.

Therefore, you need a way to "emulate" the message pushed to validate it against the consumer contract. Very likely, the API reading the queue won't have any public endpoint to hit like an REST API would, so we need a way to set up messages that would otherwise be pushed to the queue.

If you read my commented code in the event importer test class, you can see the two approaches to validate messages. One by only checking a typed object and the other one to actually execute code and by a way of mocking, get the message before it's sent to the queue and compare it with the PACT contract.

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
yanncourtel: lol you literally told me you'd prefer to use system.text in the other PR :D 
Let's make up our mind once for all about which lib we want to use. I'm all for newtonsoft though!

src/PactNet.Native/NativeMessagePactBuilder.cs
yanncourtel: As I said in my previous message, we started to ask for both ^^ Let's agree to use only newtonsoft then.

src/PactNet.Native/PactVerifier.cs
yanncourtel: I'm confused with your comment. At first, I had a MessagingPactVerifier which granted was customizing the implementation and was not really solid friendly. So I added a new method to the general verifier interface.
Are you referring to putting some kind of guard once you call one of the service provider methods?

src/PactNet.AspNetCore/PactNet.AspNetCore.csproj
yanncourtel: I think it might have been automatically added. No problem with the acceptance test.

src/PactNet.Native/PactVerifier.cs
yanncourtel: I think we should take care of it it in your oncoming PR

src/PactNet.Native/PactVerifier.cs
yanncourtel: The best practice using the framework seems to use the Uri constructor.
That would be new Uri(pactUri, messagingRoute).AbsolutePath. Not sure I'm a fan but better than using Flurl and Url.Combine.
If you have a suggestion

src/PactNet.Native/Scenarios.cs
yanncourtel: No, the Scenarios class is the inmemory storage for the scenarios and need to be public

src/PactNet.Native/Scenarios.cs
yanncourtel: I left it internal to make it testable.

src/PactNet.Native/Scenarios.cs
yanncourtel: I don't understand.
The code is scenarios.Any() == false.
Is there another way you are hinting at?

src/PactNet.Native/Scenarios.cs
yanncourtel: Good point. I used the first or default and throw on the default since it's mostly the middleware that is using it.

src/PactNet.Native/NativeMessagePactBuilder.cs
yanncourtel: Newtonsoft approach pushed

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
yanncourtel: Newtonsoft approach pushed

samples/EventImporter/Provider.Tests/TestStartup.cs
yanncourtel: This is linked to the sample app. I'll add it in another PR with the middleware unit tests. The new one should also contain the other middleware like providerstate, auth, etc.

samples/EventImporter/Provider.Tests/EventImporterProviderTests.cs
mefellows: The `AddScenario` builder actually reads quite well, it's the `WithContent` block that I'm finding hardest to follow (but probably because it's .NET specifics), but I think it does what the others do. Basically you need to get the piece of code that normally emits an event to a queue (or queue-like thing), and return it instead. 

This usually means you need to separate the code that generates the message from the bit that publishes to the queue. We care about the former. But in this case, the separation isn't there and you're intercepting the behaviour. This is probably going to be useful for those people that have good separation of concerns. It would probably be ideal to have a "here's the best practice" version of it though (doesn't need to be this PR).

For reference, this is what a message pact test looks like in 

* Java: https://docs.pact.io/implementation_guides/jvm/provider/junit/#example-of-message-test
* Ruby: https://github.com/pact-foundation/pact-message-ruby#provider
* JS: https://github.com/pact-foundation/pact-js/#consumer

@adamrodger see also this [doc](https://docs.pact.io/getting_started/how_pact_works#non-http-testing-message-pact) on "how" we test messages. It might be largely unsatisfying, but it's actually quite powerful.

I don't know the state of .NET and frameworks that wrap message queue type things, but because there is no single standard (like HTTP) we haven't taken the path of "support all of the types". This also means we can support use cases like dropping a JSON file to an S3 bucket - the expense is a fair amount of abstraction.

mefellows:
Great work Yann 👏 . My main remaining comment is docs, but that shouldn't hold up any PR.

samples/EventImporter/Provider.Tests/EventImporterProviderTests.cs
yanncourtel: Very nice comment @mefellows. 
If you look at the class EventImporterProviderTests and the methode SetupEventImporterScenarios in the sample app EventImporter I actually wrote the two approaches. Granted I didn't separate the code inside the WithContent like adding it to another class or method.

The problem I have with your example in java for instance is that you are only checking the type of the message and not the piece of code that publishes the message intercepting the message before sending it to the queue.

The WithContent method has two overloads. One with the message and one with a handle method. Would it make more sense to have the method named like this ?
- PublishMessage(dynamic message)
- PublishMessageWith(Func<dynamic> handleMessage)

yanncourtel:
> Great work Yann 👏 . My main remaining comment is docs, but that shouldn't hold up any PR.

I def have to work on the doc!

yanncourtel:
> Add the new messaging interaction model compatible with the PACT V3 specs.
> 
> This part is the provider side
> 
> On the provider side: -> add middleware to act as a messaging proxy server (so far only netstandard 2.0 compatible) -> add pact message verifying support -> add scenario builder to emulate the push of a message on the provider side (see sample app: EventImporterProviderTests).
> 
> Regarding the tests: -> add unit tests. -> add some integration tests. -> add new sample application for integrated testing with a messaging pact.

Last changes addressed the comments. For the scenario setting, we will see after the refactoring of the verifier in your oncoming PR @adamrodger 
If no more comment @adamrodger and @mefellows, we should merge it asap so I can work on integrating the other middlewares. Thanks a lot for the time commenting.

mefellows:
Makes sense to me Yann. It's still a branch, so you can continue to change the API until you're happy. Keep the flow going!

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using PactNet.AspNetCore.Messaging.Models;
+using PactNet.AspNetCore.Messaging.Options;
+using PactNet.Native;
+
+namespace PactNet.AspNetCore.Messaging
```
adamrodger, line 13: The more I've been playing around with this locally the more I'm thinking we should provide an entire server instead of only the middleware portion. It puts a lot of onus on the user to set up their own ASP.Net Core host and configure it, and all we need is a fixed endpoint so the Rust verifier can call it.

I'm not sure users will really want to do all that work of copying over all of our boilerplate, and I'm not sure it really gives any advantage to them either. Can you think of any reasons this is better than providing a single turnkey solution?

We could abstract away the entire implementation details from the user and they wouldn't even need to know a REST API had been started at all. It does feel really odd at first glance that you have to create an entire REST API to verify pub/sub messages, especially if the thing you are testing doesn't even have a REST API (like if it's a Windows service or console app that just subscribes to a queue and processes messages).

I think we could do all of that for the user instead so that they don't have to, and they'd never even need to know that there is a REST API running at all. That just becomes one of our implementation details that's hidden away.

adamrodger (changes requested):
I've made some suggestions for changing the API, and then implemented them in a branch just to validate them myself: https://github.com/pact-foundation/pact-net/tree/feature/messaging-provider-suggestions

I think the API that results from that is a bit cleaner, and also it's safer because of the type state pattern.

I've made a general suggestion about implementing the verifier as an entire turnkey solution, but I've not implemented that on my branch. I don't think it would take much work though, and it would hide away all the implementation details of us having to start an entire API to give the Rust verifier something to call.

If we did that we could also get rid of the static `Scenarios` object because then we'd have full access to DI so could just register it as a singleton in DI then inject the scenario collection into the middleware.

samples/EventImporter/Provider.Tests/EventImporterProviderTests.cs
```diff
@@ -0,0 +1,148 @@
...
+                                           "pacts",
+                                           "Event API Consumer V3 Message-Event API V3 Message.json");
+
+            //Act / Assert
+            IPactVerifier pactVerifier = new PactVerifier(config);
+            pactVerifier
+                .ServiceProvider("Event API V3 Message", this.fixture.ServerUri, options.BasePathMessage)
+                .HonoursPactWith("Event API Consumer V3 Message")
+                .FromPactFile(new FileInfo(pactPath))
+                .WithProviderMessages()
```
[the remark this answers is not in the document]
adamrodger: I've implemented this on a branch based on this MR and quite like this API:

https://github.com/pact-foundation/pact-net/tree/feature/messaging-provider-suggestions

src/PactNet.Native/MessageScenarioBuilder.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+
+namespace PactNet.Native
```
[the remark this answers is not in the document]
adamrodger: On my branch I moved all of these classes into the main `PactNet` project, so the `PactNet.AspNetCore` project no longer has any references to `PactNet.Native`.

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
yanncourtel: I believe that is how the java have done it.
That's how we've implemented it in our lib. We actually never build any service. We just register the scenarios that are "pushed" in a fake way and add it to the invokers. It's just code testing at the end.

samples/EventImporter/Provider.Tests/EventImporterProviderTests.cs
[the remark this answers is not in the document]
yanncourtel: That's a bit more user friendly indeed. I need to see it working with the injection in the middleware. (well if we end up having a middleware at all ^^)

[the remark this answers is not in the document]
yanncourtel: We will also need the overloads for the metadata. Can work yeah.

PR #329, opened 2021-10-03
chore!: Refactor verifier into type state pattern

adamrodger:
This ensures that the user can't create invalid verifier states, such as forgetting to specify provider/consumer or specifying multiple pact file sources.

The API is not backwards compatible, but existing code would only need to change if it was either invalid or didn't follow the suggested setup pattern (which would've been weird anyway).


src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <summary>
+        /// Use Basic authentication with the Pact Broker
+        /// </summary>
+        /// <param name="username">Pact broker username</param>
+        /// <param name="password">Pact broker password</param>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions BasicAuthentication(string username, string password)
+        {
+            this.verifierArgs.Add("--user",     username ?? throw new ArgumentNullException(nameof(username)));
```
yanncourtel, line 32: I wouldn't add extra space for lining up variable. I know it's for lisiblity but on different IDE, it can look different.

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+            this.verifierArgs.Add("--password", password ?? throw new ArgumentNullException(nameof(password)));
```
yanncourtel, line 33: For string, I would check for empty or white space

!string.IsNullOrWhiteSpace(password) ? password : throw new ArgumentException("password cannot be null or empty");

//For username also

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Use Token authentication with the Pact Broker
+        /// </summary>
+        /// <param name="token">Auth token</param>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions TokenAuthentication(string token)
+        {
+            this.verifierArgs.Add("--token", token ?? throw new ArgumentNullException(nameof(token)));
```
yanncourtel, line 45: Same as above for string checking

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Enable pending pacts
+        /// </summary>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions EnablePending()
+        {
+            this.verifierArgs.Add("--enable-pending", string.Empty);
```
yanncourtel, line 56: Shouldn't we add a wrapper method for adding arguments? Like AddOption("key", "value") and AddEmptyOption("key") //or just an overload

EDIT: And maybe we would factorize the checking of string arguments directly from this private method.

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        /// <summary>
+        /// Consumer tag versions to retrieve
+        /// </summary>
+        /// <param name="tags">Consumer tags</param>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions ConsumerTags(params string[] tags)
+        {
+            if (tags.Any())
```
yanncourtel, line 68: we should check if tags are null (not 100% sure with the param keyword)

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            return this;
+        }
+
+        /// <summary>
+        /// Include WIP pacts since the given date
+        /// </summary>
+        /// <param name="date">WIP cut-off date</param>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions IncludeWipPactsSince(string date)
```
yanncourtel, line 82: same comment for string verification (null or whitespace)

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <param name="tags">Optional tags to add to the verification</param>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions PublishResults(string providerVersion, params string[] tags)
+        {
+            if (string.IsNullOrWhiteSpace(providerVersion))
+            {
+                throw new ArgumentException("Can't publish verification results without the provider version being set");
+            }
+
+            this.verifierArgs.Add("--publish",          string.Empty);
```
yanncourtel, line 102: Same here, I wouldn't add extra space for visibility.

src/PactNet.Native/Verifier/NativePactVerifierPair.cs
```diff
@@ -0,0 +1,103 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Filter the interactions to only those matching the given description and/or provider state
+        /// </summary>
+        /// <param name="description">Interaction description. All interactions are verified if this is null</param>
+        /// <param name="providerState">Provider state description. All provider states are verified if this is null</param>
+        /// <returns>Fluent builder</returns>
+        public IPactVerifierPair WithFilter(string description = null, string providerState = null)
```
yanncourtel, line 48: Not sold on optional arguments. I would add overloads for this.

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace PactNet.Native.Verifier
+{
+    /// <summary>
+    /// Native pact broker options
+    /// </summary>
+    internal class NativePactBrokerOptions : IPactBrokerOptions
+    {
+        private readonly IDictionary<string, string> verifierArgs;
+
+        /// <summary>
```
yanncourtel, line 15: why not add the inherited doc tag on the implementation side? more maintainable

src/PactNet/Verifier/IPactBrokerOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <param name="tags">Consumer tags</param>
+        /// <returns>Fluent builder</returns>
+        IPactBrokerOptions ConsumerTags(params string[] tags);
+
+        /// <summary>
+        /// Include WIP pacts since the given date
+        /// </summary>
+        /// <param name="date">WIP cut-off date</param>
+        /// <returns>Fluent builder</returns>
+        IPactBrokerOptions IncludeWipPactsSince(string date); // TODO: Should this be a DateTime or DateTimeOffset so we can format it properly?
```
yanncourtel, line 43: In answer of your TODO. I think it should be and the implementation would format it correctly.

src/PactNet/Verifier/IPactVerifier.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public interface IPactVerifier
+    {
+        /// <summary>
+        /// Set the provider details
+        /// </summary>
+        /// <param name="providerName">Name of the provider</param>
+        /// <param name="pactUri">URI of the running service</param>
+        /// <returns>Fluent builder</returns>
+        IPactVerifierProvider ServiceProvider(string providerName, Uri pactUri);
```
yanncourtel, line 18: Hm, okay, I see where you are going here. Therefore, you could add a IMessagePactVerifierProvider that would eventually return another state builder.

tests/PactNet.Native.Tests/Verifier/NativePactBrokerOptionsTests.cs
```diff
@@ -0,0 +1,75 @@
+using System;
+using System.Collections.Generic;
+using FluentAssertions;
+using PactNet.Native.Verifier;
+using Xunit;
+
+namespace PactNet.Native.Tests.Verifier
+{
+    public class NativePactBrokerOptionsTests
```
yanncourtel, line 9: I would add edge cases like empty space, null arguments etc as combability tests.

For instance

[Theory]
[InlineData(null)]
[InlineData("")]
[InlineData("   ")] 
public void BasicAuthentication_Fails_WhenCalled_WithInvalidArguments(string authArg)
{
   //I verify here an exception is thrown for instance.
}

tests/PactNet.Native.Tests/Verifier/NativePactVerifierConsumerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using Xunit;
+using System.Collections.Generic;
+using FluentAssertions;
+using PactNet.Native.Verifier;
+using PactNet.Verifier;
+using Xunit.Abstractions;
+
+namespace PactNet.Native.Tests.Verifier
+{
+    public class NativePactVerifierConsumerTests
```
yanncourtel, line 12: I would add edge cases like empty space, null arguments etc as combability tests.

For instance

[Theory]
[InlineData(null)]
[InlineData("")]
[InlineData(" ")]
public void BasicAuthentication_Fails_WhenCalled_WithInvalidArguments(string authArg)
{
//I verify here an exception is thrown for instance.
}

tests/PactNet.Native.Tests/Verifier/NativePactVerifierPairTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        public NativePactVerifierPairTests(ITestOutputHelper output)
+        {
+            this.mockProvider = new Mock<IVerifierProvider>();
+
+            this.verifierArgs = new Dictionary<string, string>();
+
+            this.verifier = new NativePactVerifierPair(this.verifierArgs,
+                                                       this.mockProvider.Object,
+                                                       new PactVerifierConfig
```
yanncourtel, line 27: To fix the formatting, maybe add a variable to contain the config or like in your NativePactVerifierProviderTests test passing a newline.

yanncourtel (changes requested):

src/PactNet.Native/Verifier/NativePactVerifierConsumer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            return new NativePactVerifierPair(this.verifierArgs, new NativePactVerifier(), this.config);
+        }
+
+        /// <summary>
+        /// Use the pact broker to retrieve pact files with default options
+        /// </summary>
+        /// <param name="brokerBaseUri">Base URI for the broker</param>
+        /// <returns>Fluent builder</returns>
+        public IPactVerifierPair FromPactBroker(Uri brokerBaseUri)
```
yanncourtel, line 56: I would check if the Uri is valid (null at least)

src/PactNet/Verifier/IPactBrokerOptions.cs
```diff
@@ -0,0 +1,53 @@
+using System.Collections.Generic;
```
yanncourtel, line 1: No need for this using

src/PactNet.Native/Verifier/PactVerifier.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        public PactVerifier() : this(new PactVerifierConfig())
+        {
+        }
+
+        /// <summary>
+        /// Initialises a new instance of the <see cref="PactVerifier"/> class.
+        /// </summary>
+        /// <param name="config">Pact verifier config</param>
+        public PactVerifier(PactVerifierConfig config)
```
yanncourtel, line 26: There is a probleme here. The public constructor doesn't initialize the verifierArgs

src/PactNet.Native/Verifier/PactVerifier.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// Initialises a new instance of the <see cref="PactVerifier"/> class.
+        /// </summary>
+        /// <param name="verifierArgs">Pact verifier args</param>
+        /// <param name="config">Pact verifier config</param>
+        internal PactVerifier(IDictionary<string, string> verifierArgs, PactVerifierConfig config)
+        {
+            this.config = config;
+            this.verifierArgs = verifierArgs;
```
yanncourtel, line 39: verifier or config should not be null

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
adamrodger: Yeah fair. I think the IDE did it automatically because I never do that either.

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
adamrodger: What if the password is genuinely empty?

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
adamrodger: I don't think you get an empty array instead of null

src/PactNet.Native/Verifier/NativePactVerifierPair.cs
adamrodger: This is literally just moved as-is from where it was before without changing it

src/PactNet/Verifier/IPactBrokerOptions.cs
adamrodger: I think the problem is that it supports relative arguments like "5d" for 5 days ago, so a DateTime would prevent that. I'd have to check the CLI args though

src/PactNet.Native/Verifier/PactVerifier.cs
adamrodger: Ah that was a last minute refactor, doh

src/PactNet.Native/Verifier/PactVerifier.cs
adamrodger: This is an internal constructor so we know that these won't be null at this point

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
adamrodger: I really dislike that annotation - I want to see the docs whilst I'm in the code so that I know I'm keeping to them properly. That tag makes it far too easy to make that mistake.

tests/PactNet.Native.Tests/Verifier/NativePactBrokerOptionsTests.cs
adamrodger: Yeah I had a look and the amount of places we'd have to add this boilerplate is pretty overkill to be honest. I've added all the guard checks in by adding a specific internal extension method for it so that's probably good enough to cover this.

adamrodger:
@yanncourtel I've made all those changes and the CI should be passing now 👍

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
yanncourtel: I understand why you would want to keep it in your implementation but 1) if any change, you have to change the base symbol first and then the implementation, strenuous to maintain, 2) Your code is your documentation, in principle, you should hot have to rely on the summary doc to give you extra information.

In anyway, this is not the scope of this PR but we should address it when we review the coding and naming standard.

src/PactNet.Native/Verifier/NativePactVerifierConsumer.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <returns>Fluent builder</returns>
+        public IPactVerifierPair FromPactBroker(Uri brokerBaseUri, Action<IPactBrokerOptions> configure)
+        {
+            Guard.NotNull(brokerBaseUri, nameof(brokerBaseUri));
+            Guard.NotNull(configure, nameof(configure));
+
+            this.verifierArgs.AddOption("--broker-url", brokerBaseUri.ToString());
+
+            var options = new NativePactBrokerOptions(this.verifierArgs);
+            configure?.Invoke(options);
```
yanncourtel, line 78: I forgot since we check the configure, it needs to be mandatory so no conditional access (-> ?)

src/PactNet.Native/Verifier/NativePactBrokerOptions.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        /// <summary>
+        /// Use Token authentication with the Pact Broker
+        /// </summary>
+        /// <param name="token">Auth token</param>
+        /// <returns>Fluent builder</returns>
+        public IPactBrokerOptions TokenAuthentication(string token)
+        {
+            this.verifierArgs.AddOption("--token", token, nameof(token));
```
yanncourtel, line 45: same here (why nameof)

tests/PactNet.Native.Tests/Verifier/NativePactVerifierPairTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void WithFilter_WhenCalled_AddsFilterArgs()
+        {
+            this.verifier.WithFilter("description", "provider state");
+
+            this.CheckArgs("--filter-description", "description",
+                           "--filter-state", "provider state");
+        }
+
+        [Theory]
+        [InlineData(PactLogLevel.Trace,       "trace")]
```
yanncourtel, line 55: same here, unnecessary spacing

yanncourtel (approved):
I left some comments but not blocking for this PR.

Issue #314, opened 2021-08-12
Support Verifier Output in Provider Test Output


adamrodger:
Currently provider tests can tell you when they fail, but they give you no output from the verifier and so you are unable to determine what the actual errors are.

Adding this support requires additional work in the Rust core library, see: https://github.com/pact-foundation/pact-reference/issues/132

adamrodger:
Closed in favour of #343

mefellows:
Oh, goodness. I somehow missed this yesterday @adamrodger. It's even on the project board! I'm happy for you to kill the other issue and use this one.

See this Slack thread that prompted my to raise the issue: https://pact-foundation.slack.com/archives/C9UTHV2AD/p1638530656072500

adamrodger:
Ha, nah we'll use the other one, then it's at the top

PR #334, opened 2021-10-07
WIP feat: Add verifier logs to output (#314)

adamrodger:
Adds verifier logs from FFI to verifier output.

NOTE: This is currently broken on Windows because of an issue in the FFI: https://github.com/pact-foundation/pact-reference/issues/156. Once that issue is fixed we can bump the FFI version and this should be ready to merge

EDIT: It appears another FFI issue is also making the integration tests fail: https://github.com/pact-foundation/pact-reference/issues/157. Hopefully both issues will be fixed in the same FFI upgrade

### DIFF
--- a/build/download-native-libs.sh
+++ b/build/download-native-libs.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-FFI_VERSION="0.0.0"
+FFI_VERSION="0.0.3"
 FFI_BASE_URL="https://github.com/pact-foundation/pact-reference/releases/download/libpact_ffi-v$FFI_VERSION"
 
 GREEN="\e[32m"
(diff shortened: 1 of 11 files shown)


mefellows:
Nice!

So we still don't get the rest of the logging, but at least you'll get the verification output - which is definitely a bonus.

adamrodger:
I'm closing this MR because the FFI has gone a different way - it's moving to an approach similar to consumer tests where you have a `VerificationHandle` and you can retrieve the logs for that handle.

When we refactor to use the new FFI (which we can't just now because it doesn't support consumer version selectors in v0.1.0) then we'll implement verifier logs at that point.

adamrodger:
I just tried this on the latest FFI to be certain (v0.1.0) and it still doesn't work. You just get an empty string back from `pactffi_verifier_logs_for_provider`, presumably because the logs aren't actually being stored to the buffer.

PR #335, opened 2021-10-08
feat: provider states middleware V3 support

yanncourtel:
A PR to support provider states handling directly from a middleware in .NET core. (using a similar approach as the messaging middleware)

So far the middleware, once registered in the test dependency injection, is able to read from a set of provider states (we should maybe call them state handlers) configured by the verifier beforehand and execute a simple, parameterless, method.

Next steps:
- V3 support with parameters mapping
- Adding an action to execute provider states at set up or teardown.


yanncourtel:
@adamrodger and @mefellows this is a similar approach as the messaging middleware. It does so far execute simple methods but soon will be able to do it with parameters and also defines if the provider state (or we should name it the state handler) executes at setup or teardown.

mefellows:
Sorry, this got lost in my inbox noise Yann.

samples/PactV2/EventApi/Provider.Tests/EventAPITests.cs
```diff
@@ -0,0 +1,127 @@
...
+                .ServiceProvider("Event API", this.fixture.ServerUri)
+                .HonoursPactWith("Event API Consumer")
+                .FromPactFile(new FileInfo(pactPath))
+                .WithProviderStateUrl(new Uri(this.fixture.ServerUri, options.RouteProviderState))
+                .WithProviderStates(providerStates =>
+                {
+                    providerStates
+                        .Add(
+                            "there are events with ids '45D80D13-D5A2-48D7-8353-CBB4C0EAABF5', '83F9262F-28F1-4703-AB1A-8CFD9E8249C9' and '3E83A96B-2A0C-49B1-9959-26DF23F83AEB'",
+                            this.InsertEventsIntoDatabase, StateAction.Setup)
```
adamrodger, line 61: I'm not a big fan of the enum on the end here, especially because it's optional (and defaults to `Setup`?). I'd prefer the methods themselves to indicate what they were doing, like:

```csharp
states.AddSetup("my provider state", () => db.InsertData())
      .AddTeardown("my provider state", () => db.ClearData());
```

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            if (!(context.Request.Path.Value?.StartsWith(options.RouteProviderState) ?? false))
+            {
+                await this.next(context);
+                return;
+            }
+
+            if (context.Request.Method != HttpMethod.Post.ToString())
+            {
+                await WriteErrorInResponseAsync(context, "The provider state invocation needs to be a post method. Check on your verifier side if the provider state url is nicely configured.", HttpStatusCode.BadRequest);
```
adamrodger, line 63: This should have status code 405 - Method Not Allowed: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+                ExecuteState(providerStateBody, stateHandler);
+            }
+            context.Response.StatusCode = (int)HttpStatusCode.OK;
+
+            await WriteToResponseAsync(context, string.Empty);
+        }
+
+        private static void ExecuteState(ProviderStateInteraction providerStateBody, IStateHandler stateHandler)
+        {
+            if (providerStateBody.Params == null || !providerStateBody.Params.Any())
```
adamrodger, line 82: The design here seems a little unsafe to me - if you define a setup which takes arguments but the invocation doesn't provide any then it's ambiguous what should happen:

- Should it invoke the action anyway, even with no arguments? But then what happens to the templated parts?
- Should it error instead? I'd vote yes, given the invocation is incompatible with the defined provider state

To secure that, I'd have two different interfaces for `IStateHandler` - one which takes args and one which doesn't. That way you can't even attempt to invoke a handler which requires args from an invocation that didn't provide any (or vice versa).

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        /// <summary>
+        /// The provider state accessor
+        /// </summary>
+        private readonly IProviderStateAccessor providerStateAccessor;
+
+        ///// <summary>
+        ///// the after interaction callback
+        ///// </summary>
+        //private Dictionary<string, Action> afterInteractionCallbacks;
```
adamrodger, line 38: Not sure what this commented out code is for - does it need removing or is it unfinished?

src/PactNet.AspNetCore/ProviderState/ProviderStateOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// Defines the provider state middleware options
+    /// </summary>
+    public class ProviderStateOptions
+    {
+        /// <summary>
+        /// Route of provider state endpoint
+        /// </summary>
+        public string RouteProviderState { get; set; }
```
adamrodger, line 11: I think this should have a default value and we should have an overload in `ProviderStateMiddlewareExtensions` which just uses the defaults. That makes the simple API possible with sensible defaults:

```csharp
services.AddPactProviderState();
```

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddlewareExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        /// <summary>
+        /// Extensions method to add the provider state middleware
+        /// </summary>
+        /// <param name="builder">the builder</param>
+        /// <returns>Fluent builder</returns>
+        public static IApplicationBuilder UsePactProviderStates(this IApplicationBuilder builder)
+        {
+            IProviderStateAccessor providerStateAccessor = new ProviderStateAccessor();
```
adamrodger, line 33: Shouldn't this be registered as singleton in `AddPactProviderState` and injected via DI instead of constructed manually here?

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddlewareExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            return services;
+        }
+
+        /// <summary>
+        /// Extensions method to add the provider state middleware
+        /// </summary>
+        /// <param name="builder">the builder</param>
+        /// <returns>Fluent builder</returns>
+        public static IApplicationBuilder UsePactProviderStates(this IApplicationBuilder builder)
```
adamrodger, line 31: This method is `States` but the one above is just `State`. We should make them match.

samples/PactV3/EventApi/Consumer/DateTimeFactory.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Consumer
+{
+    public class DateTimeFactory
```
adamrodger, line 5: This is a bit weird that you overwrite the `Now` function in the tests. Usually you'd have an `IDateTimeProvider` and then you can have real and mock implementation of it.

src/PactNet.Abstractions/Verifier/IPactVerifierPair.cs
```diff
@@ -14,6 +15,12 @@ public interface IPactVerifierPair
         /// <returns>Fluent builder</returns>
         IPactVerifierPair WithProviderStateUrl(Uri providerStateUri);
 
+        /// <summary>
+        /// Adds provider states to execute
+        /// </summary>
+        /// <param name="providerStateAction"></param>
```
adamrodger, line 21: Documentation for public type

src/PactNet.Abstractions/Verifier/ProviderState/IStateHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Invoke a provider state
+        /// </summary>
+        void Execute();
+
+        /// <summary>
+        /// Invoke a provider state
+        /// </summary>
+        void Execute(IDictionary<string, string> args);
```
adamrodger, line 28: I think it the reason why splitting this into an interface with and without args is best - you can invoke this method on a handler that doesn't take arguments, or likewise you can invoke the non-argument method on a handler that requires arguments, and then it's unclear what should happen.

src/PactNet.AspNetCore/PactNet.AspNetCore.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\PactNet.Abstractions\PactNet.Abstractions.csproj" />
+    <ProjectReference Include="..\PactNet\PactNet.csproj" />
```
adamrodger, line 9: How come the ASP.Net Core project now depends on the actual implementations project? The aim of the Abstractions project is that plugins/extensions should never need to rely on the actual implementation, and that also allows all the implementation classes to be `internal` instead of `public`

src/PactNet.AspNetCore/ProviderState/ProviderStateInteraction.cs
```diff
@@ -0,0 +1,14 @@
+using System.Collections.Generic;
+using PactNet.Verifier.ProviderState;
+
+namespace PactNet.AspNetCore.ProviderState
+{
+    public class ProviderStateInteraction
```
adamrodger, line 6: Does this need to be `public`? Also we need to add docs

src/PactNet/Verifier/Messaging/MessageScenariosFactory.cs
```diff
@@ -5,13 +5,13 @@ namespace PactNet.Verifier.Messaging
     /// <summary>
     /// Message scenarios
     /// </summary>
-    public class MessageScenarios : IMessageScenarios
+    public class MessageScenariosFactory : IMessageScenariosFactory
```
adamrodger, line 8: This should be `internal` so that plugins/extensions are forced to use the abstraction

src/PactNet/Verifier/Messaging/MessagingScenarioAccessor.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace PactNet.Verifier.Messaging
+{
+    public class MessagingScenarioAccessor : IMessagingScenarioAccessor
```
adamrodger, line 5: Should be internal and have docs

src/PactNet/Verifier/Messaging/Scenario.cs
```diff
@@ -5,7 +5,7 @@ namespace PactNet.Verifier.Messaging
     /// <summary>
     /// Defines the scenario model
     /// </summary>
-    public class Scenario
+    public class Scenario : IScenario
```
adamrodger, line 8: Also internal

src/PactNet/Verifier/ProviderState/ProviderStateAccessor.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace PactNet.Verifier.ProviderState
+{
+    public class ProviderStateAccessor : IProviderStateAccessor
```
adamrodger, line 5: Should be internal and documented

src/PactNet/Verifier/ProviderState/StateHandler.cs
```diff
@@ -0,0 +1,132 @@
+using System;
+using System.Collections.Generic;
+using PactNet.Exceptions;
+
+namespace PactNet.Verifier.ProviderState
+{
+    /// <summary>
+    /// Defines the provider description model
+    /// </summary>
+    public class StateHandler : IStateHandler
```
adamrodger, line 10: Should be internal

src/PactNet/Verifier/ProviderState/StateHandler.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// The provider state description
+        /// </summary>
+        public string Description { get; }
+
+        /// <summary>
+        /// When the provider state is executed
+        /// </summary>
+        public StateAction Action { get; }
+
+        //public IStateHandler SetParams(IDictionary<string, string> args)
```
adamrodger, line 22: Remove commented code

src/PactNet/Verifier/ProviderState/StateHandler.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        /// <summary>
+        /// The invoker that will execute the provider state
+        /// </summary>
+        private readonly Action<IDictionary<string, string>> invokerWithArgs;
+
+        ///// <summary>
+        ///// Arguments
+        ///// </summary>
+        //private IDictionary<string, string> args;
```
adamrodger, line 41: Remove commented code

adamrodger (changes requested):
I've reviewed the design generally, but not reviewed the tests as you've indicated there are more to come. I'd say we're along the right track, just perhaps a few API changes to be made. 

It may have been a better idea to split the messaging middleware and provider state middleware into separate PRs though, and perhaps also the bit that moves all the sample projects around. It's made the PR really big, so I've had to wait until I could find a solid 1.5hrs to sit down and review it.

samples/PactV2/EventApi/Provider.Tests/EventAPITests.cs
yanncourtel: Good point

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddleware.cs
yanncourtel: Noted

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddleware.cs
yanncourtel: haha, leftover code

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddlewareExtensions.cs
yanncourtel: Good point

samples/PactV3/EventApi/Consumer/DateTimeFactory.cs
yanncourtel: I think this is part of the original API sample app

src/PactNet.Abstractions/Verifier/ProviderState/IStateHandler.cs
yanncourtel: Yes, I went with that design first but was stuck on the execution end. I don't remember for what reason but the final design doesn't convince me either. It works at the moment but needs refining.

src/PactNet.AspNetCore/PactNet.AspNetCore.csproj
yanncourtel: That must be a mistake. I went using interfaces to get scenarios and provider states to avoid a tight dependency to PactNet. I'll look into it.

src/PactNet/Verifier/ProviderState/StateHandler.cs
yanncourtel: Damn, a lot of leftover code :/

src/PactNet/Verifier/Messaging/MessageScenariosFactory.cs
yanncourtel: absolutely, that's the idea. I'll adjust the accessibility

yanncourtel:
> I've reviewed the design generally, but not reviewed the tests as you've indicated there are more to come. I'd say we're along the right track, just perhaps a few API changes to be made.
> 
> It may have been a better idea to split the messaging middleware and provider state middleware into separate PRs though, and perhaps also the bit that moves all the sample projects around. It's made the PR really big, so I've had to wait until I could find a solid 1.5hrs to sit down and review it.

Thanks for the review! Sorry for the big PR. Outside of the StateHandler with params, I don't see much issue.

src/PactNet.AspNetCore/ProviderState/ProviderStateMiddleware.cs
yanncourtel: Hey Adam, I'm finally having some bandwidth to work on finishing up this PR

I'm having now 2 interfaces implementing IStateHandler with different execute method. 

On the middleware, I'm having something like this : 
`        /// <summary>
        /// Execute a provider state
        /// </summary>
        /// <param name="providerStateBody">The provider state from the consumer contract</param>
        /// <param name="stateHandler">The state handler from the provider</param>
        private static void ExecuteState(ProviderStateInteraction providerStateBody, IStateHandler stateHandler)
        {
            switch (stateHandler)
            {
                case IStateHandlerWithArgs handler when providerStateBody.HasParameters():
                    handler.Execute(providerStateBody.Params);
                    break;
                case IStateHandlerWithoutArgs handlerWithoutArgs:
                    handlerWithoutArgs.Execute();
                    break;
                default:
                    throw new InvalidOperationException(
                        "The provider state in the contract is not compatible with the state definition. Please re-configure your state handler correctly.");
            }
        }`
        
        Does that sound like what you were mentioning when talking interface segregation here?

fabio-andre-rodrigues:
is this PR dead?

mefellows:
> is this PR dead?

Probably - but hopefully it gives somebody an idea of how to proceed. I'd still like to see it in Pact .NET, because making users create middleware is a pain in the butt!

PR #338, opened 2021-10-16
feat: Add consumer version selectors to verifier (closes #336)

### LINKED ISSUE #336: Add support for branches in consumer version selectors
The Pact Broker now supports branches as first class entities. You can read more about this here: https://docs.pact.io/pact_broker/branches

## Description

Please add support for specifying branch related properties in the consumer version selectors that are sent to the Pact Broker when requesting pacts to verify. Please do not do any verification of the consumer version selectors on the client side - the validation rules are subject to change. Just ensure that any error response is displayed to the user.

* For implementations that wrap the pact-ruby-standalone, update to the latest standalone version
* For implementations that use the rust implementation, update to the latest rust version.
* Add a String `branch` property to the consumer version selector (if there is a domain model for this).
* Add a Boolean `mainBranch` (or `main_branch` for snake case languages) property consumer version selector (if there is a domain model for this).
* Add a Boolean `matchingBranch` (or `matching_branch` for snake case languages) property consumer version selector (if there is a domain model for this).
* Expose and document the `branch`, `mainBranch` (or `main_branch` for snake case languages) and `matchingBranch` (or `matching_branch`) properties in the user facing API.
* Pass the `branch`, `mainBranch` and `matchingBranch` (must be camelcase) through to the relevant implementation (ruby and/or rust)

## Verifying the changes

* Publish test pacts to the test broker

```
export PACT_BROKER_BASE_URL="https://test.pact.dius.com.au"
export PACT_BROKER_USERNAME="dXfltyFMgNOFZAxr8io9wJ37iUpY42M"
export PACT_BROKER_PASSWORD="O5AIZWxelWbLvqMd8PkAVycBJh2Psyg1"

docker run --rm  \
  -e PACT_BROKER_BASE_URL  \
  -e PACT_BROKER_USERNAME  \
  -e PACT_BROKER_PASSWORD  \
  pactfoundation/pact-cli:0.50.0.14 \
  publish \
  /pact/example/pacts \
  --consumer-app-version fake-git-sha-for-demo-$(date +%s) \
  --branch main

docker run --rm  \
  -e PACT_BROKER_BASE_URL  \
  -e PACT_BROKER_USERNAME  \
  -e PACT_BROKER_PASSWORD  \
  pactfoundation/pact-cli:0.50.0.14 \
  publish \
  /pact/example/pacts \
  --consumer-app-version fake-git-sha-for-demo-$(date +%s) \
  --branch feat/x
```

* Using your pact client library, verify `{ "mainBranch": true }`
  * You should receive the main pact
* Using your pact client library, verify `{ "branch": "feat/x" }`
  * You should receive the feat/x pact
* Using your pact client library, verify `{ "matchingBranch": true }` with the provider branch set to `feat/x`
  * You should receive the feat/x pact

adamrodger:
Note: we don't need to support the `consumer` option for the selectors because that's handled outside of version selectors.

### DIFF
--- a/src/PactNet.Abstractions/Verifier/ConsumerVersionSelector.cs
+++ b/src/PactNet.Abstractions/Verifier/ConsumerVersionSelector.cs
@@ -1,0 +1,68 @@
+namespace PactNet.Verifier
+{
+    /// <summary>
+    /// Consumer version selector
+    /// </summary>
+    /// <remarks>See <see href="https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors"/></remarks>
+    public class ConsumerVersionSelector
+    {
+        /// <summary>
+        /// Select pacts from the main branch of the consumer
+        /// </summary>
+        public bool MainBranch { get; set; }
+
+        /// <summary>
+        /// Select pacts which have the same branch as the current provider branch
+        /// </summary>
+        public bool MatchingBranch { get; set; }
+
+        /// <summary>
+        /// Select pacts from the given branch
+        /// </summary>
+        public string Branch { get; set; }
+
+        /// <summary>
+        /// If the consumer doesn't have a branch which matches the <see cref="Branch"/> property, use this branch instead
+        /// </summary>
+        public string FallbackBranch { get; set; }
+
+        /// <summary>
+        /// Select pacts with the given tag
+        /// </summary>
+        /// <remarks>It is recommended that the <see cref="Branch"/> property is used instead</remarks>
+        public string Tag { get; set; }
+
+        /// <summary>
+        /// If the consumer doesn't have a tag which matches the <see cref="Tag"/> property, use this tag instead
+        /// </summary>
+        /// <remarks>It is recommended that the <see cref="FallbackBranch"/> property is used instead</remarks>
+        public string FallbackTag { get; set; }
+
+        /// <summary>
+        /// Only return consumer pacts which are deployed to an environment
+        /// </summary>
+        public bool Deployed { get; set; }
+
+        /// <summary>
+        /// Only return consumer pacts which are released and currently supported in any environment
+        /// </summary>
+        public bool Released { get; set; }
+
+        /// <summary>
+        /// Either <see cref="Deployed"/> or <see cref="Released"/> (whereas setting both of the other properties
+        /// would mean deployed AND released)
+        /// </summary>
+        public bool DeployedOrReleased { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


adamrodger:
Yep - so in PactNet you always have to provide a consumer name (at the moment, that could change in future I guess, but it's always been that way afaik) for each broker interaction.

So you don't need consumer in the version selectors because you can only ever be dealing with one consumer anyway.

If I can think of a way to auto generated the test cases for a parameterised test then perhaps that can be removed, but I've tried to think and I can't come up with a solution. It depends on the testing framework used, and I don't think we should be providing hacky extensions to test frameworks here.

bethesque:
If you don't put the consumer in, the API will return ALL the consumers though.

mefellows:
> Yep - so in PactNet you always have to provide a consumer name (at the moment, that could change in future I guess, but it's always been that way afaik) for each broker interaction.

I think we don't want this anymore, I suspect it was actually a remnant of initial versions that existed pre-broker. Having to explicitly select the consumer is not ideal as you scale a system to many consumers. 

Can we think of a reason we want to keep this behaviour? It seems a good one to break free from for the next major version.

adamrodger:
Because otherwise you just end up with one huge test that does every
consumer, branch, tag, etc all in one go. If any of that fails then it's
really difficult combing through all of the output to try and find what
happened.

Even worse, if multiple fail for different reasons then it's even more
annoying - the test keeps failing overall even as you fix one of the pacts,
so again you have to read through logs all the time to find the bits that
are failing.

If you could make that auto generate test cases then it's ok because you
could run each individual case, but I don't know of a way to do that from
within PactNet itself. At the moment I'd parameterise the test manually
(e.g. an xUnit Theory with InlineData for each consumer/branch/etc).

So, at the moment the `--consumer` flag will always be passed to the FFI,
which means it's filtered down before it gets to PactNet. It shouldn't be
receiving every pact, only the ones that match the given consumer name, as
filtered by the FFI.

mefellows:
That's definitely not consistent with the other languages (which might explain certain confusing question from the Pact .NET community). Does it force it to be the case, or just encouraged?

I agree with the pain point, but we should resolve that at the core rather than have an inconsistent API (return back structured JSON, for example). The current Ruby implementation actually does have an option to return JSON, which I use in Pact go to split into test results grouped by pact file. This is really nice, and I suggest we add that to the core also.

bethesque:
I'm still confused. If the consumer name is not passed in to which ever call gets the pacts to be verified, it will return ALL the pacts. How does not passing in the consumer allow a test to only be run for one consumer?

bethesque:
Also, does the FFI support running just a single pact by interaction ID? If it doesn't, we need to add this. The inability to run a single interaction at a time is infuriating!

PR #339, opened 2021-10-28
feat!: Remove 'HonoursPactWith' required type state for the verifier

adamrodger:
The replacement API is:

- For files - it's obviously that one consumer anyway
- For URIs - likewise
- For directories - you can filter down to consumers using the method argument
- For the broker - use the consumer property in consumer version selectors

This did require a refactor to the way we handle arguments, with a new dedicated class instead of using a dictionary. The consumer name option is repeatable, whereas using a dictionary wouldn't allow this.

The FFI is slightly awkward in this regard because sometimes options take comma-separated values (like consumer tags) and sometimes repeated arguments.


bethesque:
Thanks, that was really helpful.

Regarding: `.PublishResults(version, "prod", "staging") // <-- tags are optional`

This has raised some thoughts for me that don't need to be addressed in this PR, but I want to give you the full context while we're thinking about this interface.

Tags are always going to be supported, but their use to represent the branch is being superseded by first class support for branches. See https://docs.pact.io/blog/2021/10/08/why-we-are-getting-rid-of-tags

The data that will be sent when publishing a verification result (and pact) will now be:

* application version number - String
* branch - String (optional but very highly recommended)
* tags - Array of Strings

Coming soon:
* build URL
* commit

adamrodger:
Yeah that's a good point. That should probably work a little differently
with more fluent builder options so we can extend it in non-breaking ways
in future.

Could we raise that as an issue to support in future? Thanks

bethesque:
Copied the key points into https://github.com/pact-foundation/pact-net/issues/340

yanncourtel:
> Yeah no probs (these need to go in some actual documentation somewhere though as well):
> 
> ```cs
> IPactVerifier verifier = new PactVerifier();
> 
> // file
> verifier.ServiceProvider("My Provider", new Uri("http://localhost:5000"))
>         .WithFileSource(new FileInfo("My Provider-My Consumer.json"))
>         .Verify();
> 
> // directory
> verifier.ServiceProvider("My Provider", new Uri("http://localhost:5000"))
>         .WithDirectorySource(new DirectoryInfo("/path/to/pacts"), "Consumer 1", "Consumer 2") // <-- consumers are optional
>         .Verify();
> 
> // URI
> verifier.ServiceProvider("My Provider", new Uri("http://localhost:5000"))
>         .WithUriSource(new Uri("https://example.org/path/to/pact.json"))
>         .Verify();
> 
> // broker
> string version = Environment.GetEnvironmentVariable("GIT_SHA");
> string branch = Environment.GetEnvironmentVariable("GIT_BRANCH");
> 
> verifier.ServiceProvider("My Provider", new Uri("http://localhost:5000"))
>         .WithPactBrokerSource(new Uri("https://broker.example.org"), options =>
>         {
>             options.EnablePending()
>                    .IncludeWipPactsSince(new DateTime(2021, 02, 14))
>                    .ConsumerVersionSelectors(
>                        new ConsumerVersionSelector { MainBranch = true, Consumer = "My Consumer" },
>                        // this one is why it makes sense to have a "fallback default branch", so that we don't need to hard-code main/master per consumer here
>                        new ConsumerVersionSelector { Branch = branch, FallbackBranch = "main", Consumer = "My Consumer" },
>                        new ConsumerVersionSelector { Branch = branch, FallbackBranch = "master", Consumer = "Other Consumer" },
>                        new ConsumerVersionSelector { DeployedOrReleased = true },
>                        new ConsumerVersionSelector { Environment = "prod" }
>                        // ...and so on...
>                    )
>                    .PublishResults(version, "prod", "staging") // <-- tags are optional
>         })
>         .Verify();
> ```

Shouldn't we make it easier to filter out a specific consumer?
As a pact user, not sure I want to know I'm working with ConsumerVersionSelector to say I want a specific consumer.

How about
verifier.ServiceProvider("My Provider", new Uri("http://localhost:5000"))
.WithPactBrokerSource(new Uri("https://broker.example.org"), "Myconsumer1", "MyConsumer2", ...) //this will behind the scene create a "default" version of your option factory with just the necessary option to filter out the consumers.

OR at least
verifier.ServiceProvider("My Provider", new Uri("http://localhost:5000"))
.WithPactBrokerSource(new Uri("https://broker.example.org"), options =>
{
options.ForConsumers("Myconsumer1", "MyConsumer2", ...)
});

What do you think?

yanncourtel (commented):
Good work!

bethesque:
> .WithPactBrokerSource(new Uri("https://broker.example.org"), "Myconsumer1", "MyConsumer2", ...)

Just specifying which consumers to verify is not sufficient. The consumer versions need to be specified as well (eg. the ones from the main branch, the ones from the deployed versions etc). This is the best practice way to use pact, and we want to encourage users to use it in the recommended way. We also want the interface to be as similar as possible between implementations, so that people get a predictable experience. When each language starts customising the way they expose the options, it makes it harder for us to support and document.

adamrodger:
Can I get a thumbs up on this one so I can merge it, as there don't appear to be any major design objections? Thanks

yanncourtel (approved):

yanncourtel:
> Can I get a thumbs up on this one so I can merge it, as there don't appear to be any major design objections? Thanks

Good to go!

PR #341, opened 2021-11-13
feat: Enable the consumer of the PactVerifier to provide RequestTimeout

mortezaalizadeh:


src/PactNet/Verifier/PactVerifier.cs
```diff
@@ -34,7 +35,7 @@ public PactVerifier(PactVerifierConfig config) : this(new Dictionary<string, str
         /// </summary>
         /// <param name="verifierArgs">Pact verifier args</param>
         /// <param name="config">Pact verifier config</param>
-        internal PactVerifier(IDictionary<string, string> verifierArgs, PactVerifierConfig config)
+        public PactVerifier(IDictionary<string, string> verifierArgs, PactVerifierConfig config)
         {
```
adamrodger, line 39: This constructor is deliberately internal only. Users of the library shouldn't be able to construct it.

adamrodger (changes requested):
The new method should be added to the appropriate interface and tests should be added. I think this method belongs on the same interface as the one which allows you to set log level

src/PactNet/Verifier/PactVerifier.cs
mortezaalizadeh: @adamrodger  done

mortezaalizadeh:
@adamrodger  I added method to the interface, added test and also fixed the solution file having issue loading Consumer and Consumer.Tests project properly. Please have a look

mortezaalizadeh:
@adamrodger  we are currently using local built version of the library (V4.0), so I was wondering if you guys pushing the nuget packages somewhere we can use in our pipeline?

adamrodger:
This change will be superseded by the move to the new verifier FFI in #353. Request timeout is part of the verification options call.

Issue #327, opened 2021-09-29
Add support for publishing verification results with a branch using the pact-ruby-standalone


bethesque:
The Pact Broker now supports branches as first class entities. You can read more about this here: https://github.com/pact-foundation/docs.pact.io/blob/feat/deployments-and-releases/website/blog/2021-07-04-why-we-are-getting-rid-of-tags.md

To allow users to use this feature:

* Upgrade the pact-ruby-standalone to at least version `1.88.70`
* Add an appropriately prefixed "branch" property to the user facing interface (make it match the existing provider name, tag, and version properties)
* Pass the branch through to the pact provider verifier CLI using the parameter `--provider-version-branch`

adamrodger:
I'm not sure we're planning on doing any more Ruby-based releases (3.x and below) in favour of Rust-based releases )4.x and above) so not sure if this will get done.

adamrodger:
Yeah I'd close it. I doubt it's gonna get done

mefellows:
This seems like a fairly trivial change for a useful feature and is key to removing the dependency on tags, in favour of branches/environments/deployments. Given that most Pact .NET users would not be on the new beta, and changing to the 4.x.x line is non-trivial on behalf of the user, why would we not do this?

bethesque:
It should be a very small change and it would be great for the languages to have feature parity for the branches/envs.

YOU54F:
There was a PR done for the change in v3

https://github.com/pact-foundation/pact-net/pull/345 by @mchwalek (thank you!)

however based on this comment https://github.com/pact-foundation/pact-net/pull/345#issuecomment-1011406171 

it stalled, and was closed/continued on the v4 branch

mefellows:
> The existing API is inflexible so I can't see a good way of doing it in a non-breaking way.

Mmm quite a conundrum. A major release of the ruby based package would overlap with the new 4.x.x and cause even more confusion.

Although it's ugly, is a setter method on `PactBrokerConfig` an option e.g. `SetProviderBranch`?

The alternative is that nobody gets the feature.

adamrodger:
I'm not planning on doing any more 3.x feature releases and instead focussing on 4.x and above, so this issue can now be closed.

PR #345, opened 2022-01-03
feat: provider-version-branch option

mchwalek:
Resolves issue #327

### DIFF
--- a/PactNet.Tests/Core/PactVerifierHostConfigTests.cs
+++ b/PactNet.Tests/Core/PactVerifierHostConfigTests.cs
@@ -145,7 +145,7 @@ namespace PactNet.Tests.Core
         public void Ctor_WhenCalledWithABasicAuthenticatedBrokerConfig_SetsTheCorrectArgs()
         {
             var config = GetSubject(baseUri: new Uri("http://127.0.0.1"), 
-                brokerConfig: new PactBrokerConfig("Provider Name", "https://broker:9292/test", false, null, null, null, null),
+                brokerConfig: new PactBrokerConfig("Provider Name", "https://broker:9292/test", false, null, null, null, null, null),
                 pactBrokerUriOptions: new PactUriOptions().SetBasicAuthentication("username", "password"), 
                 providerStateSetupUri: new Uri("http://127.0.0.1/states/"));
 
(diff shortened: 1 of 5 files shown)


adamrodger (commented):
Thanks for your submission.

The code itself is fine, but I'm not sure whether it should be merged or not. The problem is that this has changed the public API (by changing a constructor) and is therefore a breaking change which requires a major release.

I don't think we're planning on doing any further major releases before v4.x though, which is a complete rewrite to use the new Rust core instead of Ruby.

mchwalek:
Hi, although I changed the public API, the change itself is backwards compatible (the added argument is optional). Therefore I think it could be a part of a minor release.

BTW the `feature/4.0.0` branch, also lacks the flag added in this PR. Should I make an equivalent PR for that branch?

mchwalek:
@adamrodger Ok, after some research, I'm convinced that my change is not backwards compatible. How about making a constructor overload instead?

adamrodger:
I think this would also need a Ruby core lib upgrade as well now that I think about it. The associated 4.x version needs an FFI upgrade, and presumably this does also. I'm not quite sure how to check what the current Ruby version actually supports though.

adamrodger:
> @adamrodger Ok, after some research, I'm convinced that my change is not backwards compatible. How about making a constructor overload instead?

I think an overload would be the only way you could do it without a breaking change, but I think the API would end up a bit of a mess just because of all the existing optional args. The new param would have to go before all of those optional args, and since it's just a string that would create a pretty error-prone API.

In fact, even that would be a breaking change. Consider this:

```csharp
public class Foo
{
    public Foo (string one, string two = null) { /* ... */ }
}

var foo = new Foo("one", "two");
```

That's fine, you've used the optional param but you didn't need to give it a name. But now someone adds a new constructor with a non-optional param before the optional ones:

```csharp
public class Foo
{
    public Foo (string one, string two = null) { /* ... */ }

    // new ctor with an extra arg before the optional
    public Foo(string one, string added, string two = null) { /* ... */ }
}

// existing code - uh-oh, doesn't this now set `added = "two"` instead of `two = "two"`?
var foo = new Foo("one", "two");
```

So existing code would change meaning, and therefore even that would be a breaking change I think. If I've got that wrong then please let me know.

mchwalek:
@adamrodger So neither a new optional param nor a constructor overload seems to be the proper solution for this. Any suggestions how should this be resolved?

adamrodger:
I'm not really sure. I don't think we want to do a major release as the Ruby core is no longer really recommended, and we're well underway with the replacement.

The existing API is inflexible so I can't see a good way of doing it in a non-breaking way.

mchwalek:
Ok, I'll close this PR then, and focus on the one for 4.0.0 release

PR #347, opened 2022-01-08
feat: add arm64 osx libs

Hoffs:
Added arm64 ffi lib for M1 Mac's

### DIFF
--- a/src/PactNet/PactNet.csproj
+++ b/src/PactNet/PactNet.csproj
@@ -22,9 +22,11 @@
     <IsWindows>False</IsWindows>
     <IsLinux>False</IsLinux>
     <IsOSX>False</IsOSX>
+    <IsArm64>False</IsArm64>
     <IsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'True'">True</IsWindows>
     <IsLinux Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'True'">True</IsLinux>
     <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'True'">True</IsOSX>
+    <IsArm64 Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture)' == 'Arm64'">True</IsArm64>
   </PropertyGroup>
 
   <ItemGroup>
@@ -46,7 +48,14 @@
       <Link>libpact_ffi.dylib</Link>
       <PackagePath>runtimes/osx-x64/native</PackagePath>
       <Pack>true</Pack>
-      <CopyToOutputDirectory Condition="'$(IsOSX)'">PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory Condition="'$(IsOSX)' == 'True' And '$(IsArm64)' == 'False'">PreserveNewest</CopyToOutputDirectory>
+      <Visible>false</Visible>
+    </Content>
+    <Content Include="$(MSBuildProjectDirectory)\..\..\build\osx\aarch64-apple-darwin\libpact_ffi.dylib">
+      <Link>libpact_ffi.dylib</Link>
+      <PackagePath>runtimes/osx-arm64/native</PackagePath>
+      <Pack>true</Pack>
+      <CopyToOutputDirectory Condition="'$(IsOSX)' == 'True' And '$(IsArm64)' == 'True'">PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
     </Content>
   </ItemGroup>
--- a/build/download-native-libs.sh
+++ b/build/download-native-libs.sh
@@ -61,3 +61,4 @@ download_native() {
 download_native "pact_ffi" "windows" "x86_64" "dll"
 download_native "libpact_ffi" "linux" "x86_64" "so"
 download_native "libpact_ffi" "osx" "x86_64" "dylib"
+download_native "libpact_ffi" "osx" "aarch64-apple-darwin" "dylib"


adamrodger:
Thanks for that. Can you also update the CI definition to make sure we're running the tests on that arch also? Hopefully that version is backwards compatible (I've not checked)

Hoffs:
Hmm, doesnt seem like github action runners support arm64 or any non x86 arch altogether.

mefellows:
Yeah, GH Actions don't yet currently support aarch64 on OSX. We had to workaround it to produce those artifacts in the first place. I can't find the issue now, but I believe they were targeting a Q1 release for it.

Hoffs:
Didn't notice tests and that updated ffi had changed handles so pact building was segfaulting. Tests now pass all green.

samples/EventImporter/Consumer.Tests/pacts/Event API Consumer V3 Message-Event API V3 Message.json
```diff
@@ -6,15 +6,24 @@
     {
       "contents": [
         {
-          "EventId": "45d80d13-d5a2-48d7-8353-cbb4c0eaabf5",
-          "EventType": "SearchView",
-          "Timestamp": "2014-06-30T01:37:41.0660548"
+          "eventId": "45d80d13-d5a2-48d7-8353-cbb4c0eaabf5",
+          "eventType": "SearchView",
+          "timestamp": "2014-06-30T01:37:41.0660548"
         }
```
adamrodger, line 12: Hmmm something has gone wrong here because the casing has changed. I'm not sure whether it was wrong previously or now though (on my phone at the moment)

samples/EventImporter/Provider/Domain/Models/Event.cs
```diff
@@ -1,16 +1,22 @@
...
 {
     public class Event
     {
+        [JsonProperty("eventId")]
         public Guid EventId { get; set; }
+
+        [JsonProperty("timestamp")]
         public DateTime Timestamp { get; set; }
+
+        [JsonProperty("eventType")]
```
adamrodger, line 14: Ah, here's the problem. Can we remove these please to keep the pacts the same?

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
```diff
@@ -99,7 +99,7 @@ private static void AddMetadataToResponse(HttpContext context, dynamic metadata)
             string metadataBase64 = Convert.ToBase64String(Encoding.UTF8.GetBytes(stringifyMetadata));
 
             context.Response.ContentType = "application/json";
-            context.Response.Headers.Add("PACT_MESSAGE_METADATA", metadataBase64);
+            context.Response.Headers.Add("pact-message-metadata", metadataBase64);
         }
```
adamrodger, line 103: Did the FFI upgrade necessitate this? If so, can you link to the relevant FFI section?

src/PactNet/PactNet.csproj
```diff
@@ -49,6 +53,13 @@
       <CopyToOutputDirectory Condition="'$(IsOSX)'">PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
     </Content>
+    <Content Include="$(MSBuildProjectDirectory)\..\..\build\osx\aarch64-apple-darwin\libpact_ffi.dylib">
+      <Link>libpact_ffi.dylib</Link>
+      <PackagePath>runtimes/osx-arm64/native</PackagePath>
+      <Pack>true</Pack>
+      <CopyToOutputDirectory Condition="'$(IsOSXArm64)'">PreserveNewest</CopyToOutputDirectory>
```
adamrodger, line 60: I think it would be better to put the condition here, so that we just set IsOSX and IsArm64, then do the full condition in this section

adamrodger (changes requested):

samples/EventImporter/Provider/Domain/Models/Event.cs
Hoffs: then the validation fails because it doesn't find the properties, as pact verifier is looking for them in camelCase.

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
Hoffs: https://github.com/pact-foundation/pact-reference/commit/b3a6f193

samples/EventImporter/Provider/Domain/Models/Event.cs
Hoffs: Also this is a change on provider side, where as consumer is generating the pact as it was previously with anonymous object where properties are CamelCased, yet the pact ends up camelCased.

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
mefellows: See also https://github.com/pact-foundation/pact-reference/tree/master/rust/pact_verifier_cli#verifying-metadata. Those docs really ought to be moved. Thanks for picking it up @Hoffs !

adamrodger:
I worked out some of the issues that you had to tackle in this PR separately for other changes, so once #350  and #351 a merged, and we can update the CI to run the M1 arch, this can be rebased and merged also 👍

samples/EventImporter/Provider/Domain/Models/Event.cs
adamrodger: This is fixed in #351 so can be stripped from this PR

samples/EventImporter/Consumer.Tests/pacts/Event API Consumer V3 Message-Event API V3 Message.json
adamrodger: Yep, see #351

src/PactNet.AspNetCore/Messaging/MessageMiddleware.cs
adamrodger: Updated in #350

Hoffs:
> @Hoffs I've merged the 0.1.6 FFI upgrade so you should be able to rebase onto that. I think only the csproj changes are needed as the rest is integrated into #350

Rebased, only csproj changes and shell script update remains.

src/PactNet/PactNet.csproj
Hoffs: From latest changes all basic conditions are separate and then only for OSX it checks 2 conditions at once.

adamrodger:
Waiting for https://github.com/actions/virtual-environments/issues/2187 so that we can add M1 CI support

adamrodger:
I didn't close this...? I think it's closed automatically because I deleted the target branch. @Hoffs please can you reopen and change the target branch to `master`? Thanks

PR #353, opened 2022-01-25
feat!: Migrate to pact verifier handles FFI

adamrodger:
The replaces the old mechanism of effectively passing command-line args to a verifier and instead moves to a more native FFI approach using proper procedural calls and handles, similar to how the mock server FFI works.

I think the FFI API will need some breaking changes as a result of some of the feedback from this work, so this is a draft PR currently. Some features currently can't be supported because of how the FFI works, such as:

- The publish options and general verification options are both mashed together in `pactffi_verifier_set_verification_options`. This means you can only set request timeouts and SSL verification options when using a Pact Broker source effectively. These should be split into separate publish and general verification options
  - Edit, see [pact-foundation/pact-reference#173](https://github.com/pact-foundation/pact-reference/pull/173)
- The `pactffi_verifier_broker_source_with_selectors` function has far too many parameters, and repeats some parameters from other functions (such as repeating `provider_name` from `pactffi_verifier_set_provider_info`)
  - Edit, see: [pact-foundation/pact-reference#174](https://github.com/pact-foundation/pact-reference/pull/174)
- Logging still doesn't work, even with the `log_to_buffer` option configured properly. Verification results always go to `stdout` regardless of what logging options you configure, which means it's still a blocker for 4.x

Edit: Marked as ready, to act a base for an eventual upgrade for the changes mention above in FFI 0.2.0. Some of the changes have already been merged but there are likely more on the way around the verifier output logging


src/PactNet/Verifier/PactBrokerPublishOptions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <summary>
+        /// Apply the configured values to the verifier
+        /// </summary>
+        public void Apply()
+        {
+            this.provider.SetVerificationOptions(true,
+                                                 this.version,
+                                                 this.buildUri,
+                                                 false, // TODO: Support disabling SSL verification
+                                                 TimeSpan.FromSeconds(5), // TODO: Support request timeouts
```
adamrodger, line 59: These will be supported in the FFI 0.2.0 upgrade

src/PactNet/Verifier/PactBrokerOptions.cs
```diff
@@ -114,20 +126,36 @@ public IPactBrokerOptions IncludeWipPactsSince(DateTime date)
...
             return this;
         }
+
+        /// <summary>
+        /// Finalise the configuration with the provider
+        /// </summary>
+        public void Apply()
+        {
+            this.provider.AddBrokerSource(brokerUri,
+                                          null, // TODO: where do we get the provider name from again?
```
adamrodger, line 149: This will be removed in the FFI 0.2.0 upgrade

src/PactNet/Verifier/PactBrokerOptions.cs
```diff
@@ -114,20 +126,36 @@ public IPactBrokerOptions IncludeWipPactsSince(DateTime date)
...
+                                          this.username,
+                                          this.password,
+                                          this.token,
+                                          this.enablePending,
+                                          this.includeWipPactsSince,
+                                          Array.Empty<string>(), // TODO: What does the provider tags arg do?
+                                          null,                  // TODO: What does the provider branch arg do?
```
adamrodger, line 156: These will be supported in the FFI 0.2.0 upgrade

adamrodger (commented):
Marked the bits of the MR with what changes will happen in the FFI 0.2.0 upgrade. I've just noticed the tests are mossing for the new publish options builder, so I'll add those now

mefellows (approved):

PR #359, opened 2022-02-06
docs: v4.x upgrade guide

adamrodger:


mefellows (approved):

mefellows:
Looks good, I suppose we should link from the main README also?

The other things that stood out to me, mostly new capabilities you're not selling!

1. You now support Message Pact 🎉 
2. Provider state handlers - no longer need to create your own custom route (I think this was added right?)
3. New matchers/generators
4. You can now use proper POSIX compliant regexes (this was a common complaint before)
5. The Windows long path issue is gone!

They might not belong there, but just went through a HL list

mefellows:
Looks good, I suppose we should link from the main README also?

The other things that stood out to me, mostly new capabilities you're not selling!

1. You now support Message Pact 🎉 
2. Provider state handlers - no longer need to create your own custom route (I think this was added right?)
3. New matchers/generators
4. You can now use proper POSIX compliant regexes (this was a common complaint before)
5. The Windows long path issue is gone!

They might not belong there, but just went through a HL list

mefellows:
> Provider state handlers - no longer need to create your own custom route (I think this was added right?)

Oh, right - [not yet](https://github.com/pact-foundation/pact-net/pull/335)

PR #360, opened 2022-02-08
ci: Create release and publish NuGet packages on tagged commits

adamrodger:
Note: I've not set the GitHub secret yet (just in case 😁 ) but I have tested that it works with the publish step commented out.

You push a tag, it creates the final NuGets, publishes them to nuget.org (using the API key held in a secret) and then creates a release on GitHub with the NuGets attached as well.


mefellows (approved):
👌 nice

PR #367, opened 2022-02-13
chore!: Upgrade to .net 6 and drop explicit .Net Framework support

adamrodger:
.Net Framework 4.6.1 and above support .Net Standard 2.0 so there's no real need to ship explicit .Net Framework support. We still run the tests on .Net Framework just to be sure that we've tested compatibility though.

This has the added bonus that the release CI can run on Ubuntu instead of Windows which means it's much much faster.


mefellows (approved):
🤘

PR #368, opened 2022-02-13
feat!: Refactor messaging API with additional features

adamrodger:
## New Features

- Support for async scenario setups
- Support custom JSON settings per scenario setup
- Simplified example with new boilerplate producer fixture (to be documented)
- More extension methods for a better API
- Fewer conflicting options in the API for configuring scenarios
- Made some types internal

## Producer Fixture Boilerplate

*Note: I'm leaving the below for posterity, but see #369 for a new embedded messaging server which makes the ASP.Net Core messaging middleware and boilerplate fixture redundant. This PR is still useful for the refactored API for specifying scenarios and the new reworked example*

The Pact messaging verification relies on having a REST API started which is called by the Rust core library with details about which messaging interaction to simulate. This is mostly handed by the `PactNet.AspNetCore.Messaging` library and an API which allows you to register handlers for each interaction.

However, that library doesn't start a full REST API, it just provides some middleware which allows you to do that. The full solution requres a real running server, but .Net doesn't have a HTTP server built in.

This begs the question "why can't the `ProviderFixture` boilerplate be shipped inside `PactNet.AspNetCore.Messaging` to stop users having to copy some boilerplate into their project?". The problem is that it relies on having the full `Microsoft.AspNetCore` package referenced in order to access classes like `WebApplication` and use Kestrel hosting to start a real API. If we added that reference then we are locked in to a yearly breaking change as and when Microsoft release a new version of ASP.Net Core.

It would mean we'd essentially have to adopt Microsoft's support policies, and also make it very difficult to backport any features. Currently we'd have to produce ASP.Net Core 3.1, 5.0 and 6.0 versions of the library because they're all still within support. That means any new features need three releases of the library, and every time Microsoft release or deprecate a version we have to respond also.

Instead, we can ship the middleware component which does the bulk of the work and then provide some boilerplate in the documentation for the user to start a real API using that middleware. This means it'll reference whichever version of ASP.Net Core they wish to use and not tie PactNet to any particular version.

I'm going to look into using https://github.com/unosquare/embedio to embed the tiny REST API directly inside a `PactNet.Messaging` library so that the user doesn't have to do anything. This is similar to what PactGo does, because Go has a HTTP server built into the standard library, which makes messaging pacts as seamless as 'regular' sync pacts.


mefellows:
This looks great. I'm not in a good position to review this given the .NET-isms in it, but removing the boilerplate will go a long way to improving the DX!

adamrodger:
See #369 for replacing the boilerplate messaging verifier fixture with a new built in messaging server, which deprcates the `PactNet.AspNetCore` package (before it's even been released, so I guess deprecate is the wrong word?) and removes the need for the user to deal with the messaging verifier at all.

I'll leave it as a part of the PR, but it probably doesn't need much review given it's essentially DOA.

adamrodger:
Merging due to the improved API, and Matt has given it a once over 👍

Issue #382, opened 2022-04-06
pact deserialization fails when Json SnakeCaseNamingStrategy is used


mornemaritz-bmw:
When using custom `JsonSerializerSettings` with a `ContractResolver` that has a `NamingStrategy` set to `SnakeCaseNamingStrategy`, the `IMessageVerifier.Verify<T>` operation fails to correctly deserialize properties of the target object that include underscores.

The following setup for a Pact will result in a failure to deserialize any properties on the target object that map to a property in the JSON which contains an underscore.
```
            IMessagePactV3 v3 = MessagePact.V3("Extended Stock Event Consumer", "Extended Stock Event Producer", new PactConfig
            {
                PactDir = "../../../pacts/",
                DefaultJsonSettings = new JsonSerializerSettings
                {
                    ContractResolver = new DefaultContractResolver { NamingStrategy = new SnakeCaseNamingStrategy() },
                },
                Outputters = new[]
                {
                    new XUnitOutput(output)
                }
            });

            this.messagePact = v3.UsingNativeBackend();
```

While investigating the issue, I found that the calls to [`JsonConvert.DeserializeObject<T>(targetObject)`](https://github.com/pact-foundation/pact-net/blob/0ea980d20f1ea7437f132145a4e7944417ae04b6/src/PactNet/ConfiguredMessageVerifier.cs#L85) in [`ConfiguredMessageVerifier.MessageReified<T>()`](https://github.com/pact-foundation/pact-net/blob/0ea980d20f1ea7437f132145a4e7944417ae04b6/src/PactNet/ConfiguredMessageVerifier.cs#L80) do not use the `DefaultJsonSettings` that have been set on the [`pactConfig`](https://github.com/pact-foundation/pact-net/blob/0ea980d20f1ea7437f132145a4e7944417ae04b6/src/PactNet/ConfiguredMessageVerifier.cs#L18)

I did find that simply specifying the configured settings as the relevant parameter in call to `DeserializeObject<T>` fixed the issue: `JsonConvert.DeserializeObject<T>(targetObject, config.DefaultJsonSettings)`

mornemaritz-bmw:
I can confirm that the issue has been fixed. Thank you for your quick action on this.

PR #383, opened 2022-04-06
fix: json naming strategy support

mornemaritz-bmw:
This is a suggested fix for https://github.com/pact-foundation/pact-net/issues/382

### DIFF
--- a/samples/Messaging/Consumer.Tests/JsonNamingStrategyTests.cs
+++ b/samples/Messaging/Consumer.Tests/JsonNamingStrategyTests.cs
@@ -1,0 +1,65 @@
+﻿using System.Collections.Generic;
+using FluentAssertions;
+using FluentAssertions.Extensions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using PactNet;
+using PactNet.Matchers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Consumer.Tests
+{
+    public class JsonNamingStrategyTests
+    {
+        private readonly JsonSerializerSettings jsonSerializerSettings;
+        private readonly IMessagePactBuilderV3 messagePact;
+
+        public JsonNamingStrategyTests(ITestOutputHelper output)
+        {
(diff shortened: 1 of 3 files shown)


adamrodger:
The fix itself is fine, it just needs the unit tests updating. There are similar unit tests for making sure user-provided settings are used in the non-messaging unit tests if you need a pointer.

src/PactNet/ConfiguredMessageVerifier.cs
```diff
@@ -80,9 +80,9 @@ public async Task VerifyAsync<T>(Func<T, Task> handler)
         private T MessageReified<T>()
         {
             string reified = this.server.Reify(this.message);
-            NativeMessage content = JsonConvert.DeserializeObject<NativeMessage>(reified);
+            NativeMessage content = JsonConvert.DeserializeObject<NativeMessage>(reified, config.DefaultJsonSettings);
```
adamrodger, line 83: I don't think this line needs the settings adding. This is the response from the Rust FFI which wraps the message, not the actual message itself. It's the reified version below that needs the settings to be applied.

samples/Messaging/Consumer.Tests/JsonNamingStrategyTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using PactNet;
+using PactNet.Matchers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Consumer.Tests
+{
+    public class JsonNamingStrategyTests
+    {
+        private readonly JsonSerializerSettings jsonSerializerSettings;
```
adamrodger, line 15: As mentioned in the MR comment, this can be covered by unit tests instead of integration tests, which is a neater solution. Check out the non-messaging unit tests for pointers.

adamrodger (changes requested):

adamrodger:
@mornemaritz-bmw You were right - I went to find the tests to help you out and I found there were no tests for this.

I added them in https://github.com/pact-foundation/pact-net/commit/cb7693dd4e8dcc4dee301d1a7617b8ed54c93425 which also added the small fix required to solve the original issue, so this PR no longer needs merging. Thank you for your contribution!

Issue #392, opened 2022-05-26
Support Publishing Results Using URI Sources


adamrodger:
## Existing State

Currently, PactNet only supports publishing verification results using the Broker source, like this:

```csharp
string version = Environment.GetEnvironmentVariable("GIT_SHA");
string branch = Environment.GetEnvironmentVariable("GIT_BRANCH");
string username = Environment.GetEnvironmentVariable("BROKER_USERNAME");
string password = Environment.GetEnvironmentVariable("BROKER_PASSWORD");

verifier.ServiceProvider("My API", this.fixture.ServerUri)
        .WithPactBrokerSource(new Uri("https://broker.example.org"), options =>
        {
            options.EnablePending()
                   .ConsumerVersionSelectors(new ConsumerVersionSelector { MainBranch = true })
                   .PublishResults(version, publish => publish.ProviderBranch(branch)); // 👈 enable publishing results
        })
        .Verify();
```

## Proposed API

It should be possible to publish verification results to a Pact Broker when using a URI source, as recommended in some of the [Pact CI/CD setup guides](https://docs.pact.io/pact_nirvana). In these flows, a verifier is triggered from a webhook with the URI of the pact that has changed (instead of all pacts). The verifier should then pull the pact from this URI, verify just this one pact, then publish results back to the broker.

The API would be something like:

```csharp
Uri uri = new Uri(Environment.GetEnvironmentVariable("PACT_URI")); // 👈 from the webhook

verifier.ServiceProvider("My API", this.fixture.ServerUri)
        .WithUriSource(uri, options =>
        {
            options.BasicAuthentication(username, password)
                   .PublishResults(version, publish => publish.ProviderBranch(branch)); // 👈 new
        })
        .Verify();
```

## Open Questions

- How is the broker URI discovered?
    - Is it inferred from the webhook URI itself?
    - Is that even safe?
    - What if you retrieve the pact from any old URI and then try to publish results as if it's a broker?
    - Should we force people to provide the broker URI instead? That seems the safest solution
    - **Answer**: The pact file itself will contain a link to publish verification results if it was retrieved from a Pact Broker. Thus, if the link is missing then the publish option shouldn't do anything, and if it is present then publshing can be safely achieved.
- What support is available in the FFI for this?
    - **Answer**: None currently. See the tracking issue: https://github.com/pact-foundation/pact-reference/issues/199
[...]

PR #393, opened 2022-05-26
feat: Allow publishing verification results for URI sources (#392)

adamrodger:
See the linked issue #392 for additional information

Draft currently as it's unclear where the broker URI is going to come from. Perhaps this is built into the FFI, but that doesn't seem safe to me.


bethesque:
See https://github.com/pact-foundation/pact-net/issues/392#issuecomment-1139328962 for where the verification results URL comes from.

adamrodger:
This can be merged when it has a related FFI update which includes feature https://github.com/pact-foundation/pact-reference/issues/199

The PactNet side shouldn't need changing further because the actual change is on the FFI side. As long as we call the `pactffi_set_publish_options` function with a URL source (which this PR adds) then the FFI will be responsible for publishing the results internally.

Issue #391, opened 2022-05-26
Support Non-JSON Text Bodies


adamrodger:
PactNet should support request and response bodies which are textual but not JSON, such as CSV or just plain text to support [ASP.Net Core health checks](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks), for example.

## Current Problem

The only available option is `WithJsonBody`, like this:

```csharp
pact.UponReceiving("a request to retrieve all events")
        .WithRequest(HttpMethod.Get, "/events")
    .WillRespond()
        .WithStatus(HttpStatusCode.OK)
        .WithJsonBody(Match.MinType(new
        {
            eventId = Match.Type("3E83A96B-2A0C-49B1-9959-26DF23F83AEB"),
            timestamp = Match.Type("2014-06-30T01:38:00.8518952"),
            eventType = Match.Regex("SearchView", "SearchView|DetailsView")
        }, 1));
```

if you use that to specify any kind of fixed string then the string is serialised as JSON, and thus wrapped in quotes with some escaping performed (such as escaping new lines) in the resultant pact file.

## Proposed Solution

We should allow a flexible escape hatch for when the user just needs to specify some text as-is without any interference, any matching rules or anything else, and also specify the content type.

The API would be something like:

```csharp
pact.UponReceiving("a request for a health check")
        .WithRequest(HttpMethod.Get, "/healthz")
    .WillRespond()
        .WithStatus(HttpStatusCode.OK)
        .WithTextBody("Healthy", "text/plain");

pact.UponReceiving("a request for some CSV")
        .WithRequest(HttpMethod.Get, "/export")
    .WillRespond()
        .WithStatus(HttpStatusCode.OK)
        .WithTextBody("header1,header2\nvalue1,value2", "text/csv");
```

PR #395, opened 2022-05-27
feat: Support non-JSON request and response bodies (#391)

adamrodger:
Quite a simple change to just pass through some text as-is to use as the request or response body. The existing JSON methods can just run the serialiser and then pass the string result to the plain text version, so this is totally backwards compatible.

It would've been nice to use `Span<char>` instead of string, but since we're on `netstandard2.0` then that's not available.


mefellows (approved):
LGTM

PR #422, opened 2022-10-31
Support provider state teardown

sprengo:


src/PactNet.Abstractions/Verifier/IPactVerifierSource.cs
```diff
@@ -7,13 +7,14 @@ namespace PactNet.Verifier
...
+    {
         /// <summary>
         /// Set up the provider state setup URL so the service can configure states
         /// </summary>
-        /// <param name="providerStateUri">Provider state setup URI</param>
+        /// <param name="providerStateUri">Provider state setup URI</param>
+        /// <param name="teardown">Whether to invoke a teardown request to the provider state endpoint after each interaction</param>
         /// <returns>Fluent builder</returns>
-        IPactVerifierSource WithProviderStateUrl(Uri providerStateUri);
+        IPactVerifierSource WithProviderStateUrl(Uri providerStateUri, bool teardown = false);
```
adamrodger, line 17: This is a breaking API change, when though it's optional, so needs to be a separate overload instead of an optional parameter

src/PactNet.Abstractions/Verifier/IPactVerifierSource.cs
```diff
@@ -7,13 +7,14 @@ namespace PactNet.Verifier
     /// Configured pact verifier source
     /// </summary>
     public interface IPactVerifierSource
-    {
+    {
```
adamrodger, line 10: Are there some line ending changes here or something? Unchanged lines appear to be marked as change

adamrodger (changes requested):

adamrodger:
Please could you also change the commit message to follow the conventional commits style?

adamrodger:
I'm wondering what to do about the API for this because there's another bool parameter on the FFI call, and we should probably add support for that at the same time.

If we just add the one parameter then we'll make adding a second parameter awkward, because that would be (Uri, bool, bool). If you just wanted to set the second bool without the first then you couldn't, because they would both have the signature (Uri, bool).

We should perhaps make those two parameters enums instead of using bools there so that we can add the appropriate overloads and communicate the intent of those parameters more clearly.

sprengo:
Any suggestions how to name the enums?

ProviderStateInvocationBehavior { SetupOnly, WithTeardown }
ProviderStateTransmissionType { QueryParams, Body }

philek:
I would go the simplest route with only one overload of (Uri, bool, bool). The options are simple and there's no need to complicate it with enums and additional overloads. At most add default values so you can only override one.

adamrodger:
Provider state configuration support added in https://github.com/pact-foundation/pact-net/commit/c2ec3cc4b1a4f65b99aaaa7625af49bf6fbc4700

Issue #421, opened 2022-10-20
Why is teardown flag not supported for WithProviderStateUrl?


sprengo:
Why not just add the teardown flag to `WithProviderStateUrl` method of the `IPactVerifierSource` interface and pass it to the `SetProviderState` of the `IVerifierProvider` in the corresponding implementation?

adamrodger:
Added in https://github.com/pact-foundation/pact-net/commit/c2ec3cc4b1a4f65b99aaaa7625af49bf6fbc4700

PR #439, opened 2023-01-25
feat: Support provider state teardown requests (#421)

philek:
Adding support for provider state teardown requests and state sent as query params.

This is based on #422.


adamrodger:
Provider state configuration support added in https://github.com/pact-foundation/pact-net/commit/c2ec3cc4b1a4f65b99aaaa7625af49bf6fbc4700

PR #446, opened 2023-03-22
feat: xunit output adapter package

riezebosch:
I've moved the xunit adapter into a separate project in order to have you publish it as a separate package, since we'll all be using it all over again


riezebosch:
I'm struggling to get a proper name for the extension method:

```csharp
new PactVerifierConfig
{
    Outputters = new[]
    {
        output.FromPact(),
        output.ToPact(),
        output.AsPact(),
        output.AsPactOutput(),
        output.AsPactOutputter(),
    }
});
```

adamrodger:
I think I prefer that as a class instead of an extension method.

The extension method will always show up in intellisense lists, and it's really annoying when libraries are far too liberal with exposing extension methods for common things.

I think a simple XUnitOutput class which implements IOutput and takes the xUnit output object in the ctor is absolutely fine.

I'm struggling with the idea of publishing this as a separate package though. Like it's literally going to be two lines of code, but we're baking in things like minimal xUnit and .Net versions in order to use those two lines. Just seems a bit overkill for what it is.

riezebosch:
It is possible target the lowest version of `xunit.abstractions` possible (from 2015!): 

```
    <ItemGroup>
      <PackageReference Include="xunit.abstractions" Version="2.0.0" />
    </ItemGroup>
```

And also target `netstandard2.0` with [version support for](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0):

.NET and .NET Core | 2.0, 2.1, 2.2, 3.0, 3.1, 5.0, 6.0, 7.0
-- | --
.NET Framework 1 | 4.6.1 2, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8, 4.8.1

Personally, I do like the extension method since it targets the `ITestOutputHelper` specifically and will only show up there and for its expressiveness. But for sure, that's a matter of taste.

adamrodger:
I just realised this is a duplicate of #372 so closing in favour of that earlier MR

Issue #410, opened 2022-07-01
Need support for multipart/form-data while writing interaction


prasadsolanki:
I am writing a contract test for a post method where I need to upload a file in request Body as 'form-data'. I didn't see any multipart content specific function  in PactNet4. All the available functions expecting JSON body
![image](https://user-images.githubusercontent.com/21264793/176871229-3c68cbd0-b34e-4579-adcb-c138a3aff119.png)

adamrodger:
That's correct; we currently support JSON bodies only.

prasadsolanki:
Is there any ETA for this? If its going to be too far, I can switch to other language e.g. PactJS. But again can someone confirm if Pact JS supports this multipart/form-data file upload?

mefellows:
Yes Pact JS (the beta) supports multi-part, but I don't see how that would help. You can't write unit tests for a .NET language in JS.

adamrodger:
Note to future self - The FFI has a method for adding multipart messages named `pactffi_with_multipart_file`

mefellows:
I've just added a label "smartbear supported" to the ticket, which will create an internal tracking issue in PactFlow's Jira (see that comment above). We will use this to prioritise and assign a team member to this task.

All activity will be public on this ticket. For now, sit tight and we'll update this ticket once we have more information on the next steps.

mefellows:
Thanks Adam, will do.

Inksprout:
Hi @adamrodger, I'm looking into implementing this now. I'll propose an API for it here and we can discuss before I get deep into implementing it. Let me know if you would rather I raise the API proposal elsewhere.

The fields needed for the pactffi_with_multipart_file Rust core method are
- Interaction: this is already available in the existing code
- part: this is already available in the existing code
- content_type: We can hard code this as we've done with the  'WithJsonBody'  Request builder, since this api will only be for multipart/form-data content type. (that's my understanding of the requirements at least)
- file: User will need to pass in the absolute file path, as the current directory when the tests run could be inside bin/Debug.
- part_name: This is the delimiter in the body content that marks the different parts. There is no matching performed on this string as part of pact file verification so it is needed to call the Rust library but can be an arbitrary string. For this reason we can set a sensible default as well.

So I'm proposing the response builder interface would look like this, with the other fields added programmatically: 

```
IRequestBuilderV3 IRequestBuilderV3.withMultipartFileUpload(string filePath, string partName = "partDelimiter")
[...]

PR #447, opened 2023-03-23
Feat multipart file upload support

Inksprout:
This PR aims to add partial support for Multipart/form-data requests to be added to the Pact File. 
It introduces a WithMultipartSingleFileUpload() method that allows the user to to specify a request of multipart/form-data in which a single part is uploaded, which is a file.

This functionality is a common use case and uses the underlying Rust FFI method designed for this express purpose.
Note this update does not enable Pact-Net to fully support all types of multipart/form-data requests as it does not support requests with multiple parts included.

see issue: https://github.com/pact-foundation/pact-net/issues/410

TODO

- [x] Add documentation for the new method


src/PactNet.Abstractions/IRequestBuilder.cs
```diff
@@ -204,7 +204,14 @@ public interface IRequestBuilderV3
...
         /// <returns>Fluent builder</returns>
-        IRequestBuilderV3 WithBody(string body, string contentType);
+        IRequestBuilderV3 WithBody(string body, string contentType);
+
+        /// <summary>
+        /// A Multipart body containing a single part, which is an uploaded file
+        /// </summary>
+        /// <param name="body">Request body</param>
+        /// <returns>Fluent builder</returns>
+        IRequestBuilderV3 WithMultipartSingleFileUpload(string filePath);
```
mefellows, line 214: @uglyog is this supported on the V2 interface as well? I'm assuming it is, because there is nothing inherently new about the spec it writes out?

src/PactNet/Drivers/IHttpInteractionDriver.cs
```diff
@@ -54,6 +54,15 @@ internal interface IHttpInteractionDriver : IProviderStateDriver
         /// </summary>
         /// <param name="contentType">Context type</param>
         /// <param name="body">Serialised body</param>
-        void WithResponseBody(string contentType, string body);
+        void WithResponseBody(string contentType, string body);
+
+
```
mefellows, line 59: Looks like an additional white space here - is there some common way to format C# code?

src/PactNet.Abstractions/IRequestBuilder.cs
rholshausen: It uses the content type matcher, which V3.

src/PactNet/Drivers/InteropActionExtensions.cs
```diff
@@ -19,5 +21,14 @@ public static void CheckInteropSuccess(this bool success)
                 throw new PactFailureException("Unable to perform the given action. The interop call indicated failure");
             }
         }
+
+        public static void CheckInteropSuccess(this StringResult success)
+        {
+            if (success.tag!=StringResult.Tag.StringResult_Ok)
```
mefellows, line 27: similar formatting question - based on what I see elsewhere there is usually space before/after `!=`.

src/PactNet/Interop/StringResult.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Runtime.InteropServices;
+namespace PactNet.Interop
+{
+    /// <summary>
+    /// Represents the result of a string operation.
```
mefellows, line 6: My understanding is that this enum stores a "result" of an operation (not necessarily a string operation). The result is either a success, and there are potential messages associated with the OK state or Failed state.

@uglyog could you please clarify?

src/PactNet/RequestBuilder.cs
```diff
@@ -390,8 +398,20 @@ internal RequestBuilder WithJsonBody(dynamic body, JsonSerializerSettings settin
         {
             string serialised = JsonConvert.SerializeObject(body, settings);
             return this.WithBody(serialised, contentType);
+        }
+
+        /// <summary>
+        /// Set a body which is multipart/form-data but contains only one part, which is a file upload
```
mefellows, line 404: Should this description be the same as the overloaded one above (I think this version is the better description).

I'm also assuming the case of the parameter descriptions should have the first word capitalised? (ditto above method)

mefellows (changes requested):
Looks good so far. I only have bike shed comments to offer at this stage :)

src/PactNet/Interop/StringResult.cs
rholshausen: That comment is wrong, it can be any operation. Also, I don't think the .Net layer need be concerned with the success value, only if there is an error, that needs to be presented to the user.

src/PactNet.Abstractions/IRequestBuilder.cs
mefellows: ok cool, we'll leave it out then. Thanks.

src/PactNet/Interop/StringResult.cs
mefellows: Thanks. I couldn't see when the success string might be populated.

adamrodger:
Nice, I'll give this a review later on 👍

src/PactNet.Abstractions/IRequestBuilder.cs
```diff
@@ -204,7 +204,14 @@ public interface IRequestBuilderV3
...
         /// <returns>Fluent builder</returns>
-        IRequestBuilderV3 WithBody(string body, string contentType);
+        IRequestBuilderV3 WithBody(string body, string contentType);
+
+        /// <summary>
+        /// A Multipart body containing a single part, which is an uploaded file
+        /// </summary>
+        /// <param name="filePath">The absolute path of the file being uploaded</param>
+        /// <returns>Fluent builder</returns>
+        IRequestBuilderV3 WithMultipartSingleFileUpload(string filePath);
```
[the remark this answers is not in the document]
adamrodger: The name of this method is a bit confusing to me. Like I get why it's called that but it contains both "multi" and "single", which are opposites of each other. We should probably think of a shorter name that's a bit easier to understand.

src/PactNet/Interop/NativeInterop.cs
```diff
@@ -61,7 +61,10 @@ internal static class NativeInterop
         public static extern bool ResponseStatus(InteractionHandle interaction, ushort status);
 
         [DllImport(DllName, EntryPoint = "pactffi_with_body")]
-        public static extern bool WithBody(InteractionHandle interaction, InteractionPart part, string contentType, string body);
+        public static extern bool WithBody(InteractionHandle interaction, InteractionPart part, string contentType, string body);
+
+        [DllImport(DllName, EntryPoint = "pactffi_with_multipart_file")]
+        public static extern StringResult WithMultipartSingleFileUpload(InteractionHandle interaction, InteractionPart part, string contentType, string filePath,  string mimePartName );
```
adamrodger, line 67: Errant space on the end here

src/PactNet/Drivers/InteropActionExtensions.cs
```diff
@@ -19,5 +21,14 @@ public static void CheckInteropSuccess(this bool success)
                 throw new PactFailureException("Unable to perform the given action. The interop call indicated failure");
             }
         }
+
+        public static void CheckInteropSuccess(this StringResult success)
```
adamrodger, line 25: Missing docs

src/PactNet/Interop/StringResult.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
```
adamrodger, line 1: Some formatting issues in this file:

- Needs a blank line after the imports
- Names contain underscores, which doesn't match C# naming conventions
- Fields named `0` aren't descriptive

src/PactNet/RequestBuilder.cs
```diff
@@ -241,8 +241,16 @@ IRequestBuilderV3 IRequestBuilderV3.WithJsonBody(dynamic body, JsonSerializerSet
...
-
+            => this.WithJsonBody(body, settings, contentType);
+
+        /// <summary>
+        /// Set a body which is multipart/form-data but contains only one part, which is a file upload
+        /// </summary>
+        /// <param name="filePath">Path to the file being uploaded</param>
+        /// <returns>Fluent builder</returns>
+        IRequestBuilderV3 IRequestBuilderV3.WithMultipartSingleFileUpload(string filePath)
+            => this.WithMultipartSingleFileUpload(filePath, "multipart/form-data", "file");
```
adamrodger, line 252: You can't hard-code `"file"` here because that influences the name of the part on the provider side, which might be important.

src/PactNet/RequestBuilder.cs
```diff
@@ -241,8 +241,16 @@ IRequestBuilderV3 IRequestBuilderV3.WithJsonBody(dynamic body, JsonSerializerSet
         /// <param name="contentType">Content type override</param>
         /// <returns>Fluent builder</returns>
         IRequestBuilderV3 IRequestBuilderV3.WithJsonBody(dynamic body, JsonSerializerSettings settings, string contentType)
-            => this.WithJsonBody(body, settings, contentType);
-
+            => this.WithJsonBody(body, settings, contentType);
```
adamrodger, line 244: I think there are generally weird line-ending problems throughout the PR because most changed files say that blank lines have been deleted and added back again. We should check to make sure the line endings are normalised.

tests/PactNet.Tests/Drivers/FfiIntegrationTests.cs
```diff
@@ -24,28 +27,29 @@ public FfiIntegrationTests(ITestOutputHelper output)
             this.output = output;
 
             NativeInterop.LogToBuffer(LevelFilter.Trace);
-        }
-
+        }
+
         [Fact]
-        public async Task HttpInteraction_v3_CreatesPactFile()
```
adamrodger, line 30: The line-endings problem has really confused the diff here. I can't really follow this as-is because it thinks an existing large integration test was rewritten instead of a new test added.

tests/PactNet.Tests/data/v3-server-integration-MultipartFormDataBody.json
```diff
@@ -0,0 +1,69 @@
...
...NxdWtmYmc2WjJTDQpDb250ZW50LURpc3Bvc2l0aW9uOiBmb3JtLWRhdGE7IG5hbWU9ImZpbGUiOyBmaWxlbmFtZT0idGVzdF9maWxlLmpwZWciDQpDb250ZW50LVR5cGU6IGltYWdlL2pwZWcNCg0K/9j/4AAQSkZJRgABAQAAAQABAAD/4gHYSUNDX1BST0ZJTEUAAQEAAAHIAAAAAAQwAABtbnRyUkdCIFhZWiAAAAAAAAAAAAAAAABhY3NwAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAA9tYAAQAAAADTLQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAlkZXNjAAAA8AAAACRyWFlaAAABFAAAABRnWFlaAAABKAAAABRiWFlaAAABPAAAABR3dHB0AAABUAAAABRyVFJDAAABZAAAAChnVFJDAAABZAAAAChiVFJDAAABZAAAAChjcHJ0AAABjAAAADxtbHVjAAAAAAAAAAEAAAAMZW5VUwAAAAgAAAAcAHMAUgBHAEJYWVogAAAAAAAAb6IAADj1AAADkFhZWiAAAAAAAABimQAAt4UAABjaWFlaIAAAAAAAACSgAAAPhAAAts9YWVogAAAAAAAA9tYAAQAAAADTLXBhcmEAAAAAAAQAAAACZmYAAPKnAAANWQAAE9AAAApbAAAAAAAAAABtbHVjAAAAAAAAAAEAAAAMZW5VUwAAACAAAAAcAEcAbwBvAGcAbABlACAASQBuAGMALgAgADIAMAAxADb/2wBDAAEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQH/2wBDAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQH/wAARCAABAAEDAREAAhEBAxEB/8QAFAABAAAAAAAAAAAAAAAAAAAABv/EABQQAQAAAAAAAAAAAAAAAAAAAAD/xAAVAQEBAAAAAAAAAAAAAAAAAAAJCv/EABQRAQAAAAAAAAAAAAAAAAAAAAD/2gAMAwEAAhEDEQA/AC6lAjD/2Q0KLS1MQ1NqcTNxdWtmYmc2WjJTLS0NCg==",
```
adamrodger, line 14: Oh, this isn't what I expected to see... How can we validate that the body is actually a multi-part one in the correct format if the entire body is base64 encoded? That's made the interaction pretty opaque, which will make it hard to debug if the verifier fails.

adamrodger: I've decoded this body and I'm confused by something. The body is:

```
--LCSjq3qukfbg6Z2S
Content-Disposition: form-data; name="file"; filename="test_file.jpeg"
Content-Type: image/jpeg

<binary data>
--LCSjq3qukfbg6Z2S--
```

but where does the `Content-Type: image/jpeg` come from? The public API doesn't allow you to specify the content type of a part as far as I can see. Is that being auto-detected or something? Either way, you should probably be able to set it yourself.

adamrodger (changes requested):
In addition to the comments, I'd like to see an additional test for the verifier side so that we can confirm the entire round-trip works.

tests/PactNet.Tests/Drivers/FfiIntegrationTests.cs
Inksprout: Sorry Adam, I think I have fixed the issue now

src/PactNet.Abstractions/IRequestBuilder.cs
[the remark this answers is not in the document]
Inksprout: I was going off the name @mefellows suggested on the issue thread, will come up with something briefer  😆

tests/PactNet.Tests/data/v3-server-integration-MultipartFormDataBody.json
Inksprout: This is a similar issue to the "file" one you mentioned, I have updated the code to include this as a parameter the user sets. There is a limitation here I am planning to document however, the content type matching in the Rust library uses the Shared MIME-info Database (https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html) to do the binary comparison. This library is GNU licensed, so we can't embed it in the framework, it has to be installed in the system. It is not supported in windows, so while content matching will work as expected in a Unix type build pipeline, the content type will only come through as 'application/octet-stream' when using windows. You can see the test I added reflects this, as I have to check and set the content type appropriately so that test can pass in each of the test environments the repo has.

adamrodger: Oh wow, that's quite the gotcha I wasn't aware of. So running the same test on Windows and Linux will produce different pact files? That doesn't sound good.

What if you run it on Linux with and without the mime detection library installed? Does it error or give you different results also?

Inksprout: I had a chat with Ron who is the main maintainer for the Rust core about this. He recommends skipping the tests on windows rather than incorporating an OS check into them, as this should reduce confusion and is how some of the other pact libraries have handled it.
I've updated my tests to reflect this. I have also updated the docs with details on how to use the method and it's limitations. Take a look, I think these should help prevent too many issues arising. Happy to add more detail there too.

mefellows: Just linking to this related issue: https://github.com/pact-foundation/pact-reference/issues/171

adamrodger: I'm not really willing to skip the tests on Windows for a .Net library.

Windows is the main development OS for .Net by far, but obviously they also need to work in other OSs as people increasingly run in dev containers or run their CI on different OS.

A pretty typical workflow, and the one everyone uses at my work, is for people to develop on Windows but run CI and production in Linux containers. We need the same tests to produce the same results in that situation.

I honestly think this change is blocked into we get an FFI behaviour change. Fortunately the issue is linked now so we can track it.

mefellows: > A pretty typical workflow, and the one everyone uses at my work, is for people to develop on Windows but run CI and production in Linux containers. We need the same tests to produce the same results in that situation.

If I understand correctly, the issue here will be that the CI will publish a contract with the correct MIME type e.g. `image/jpeg` and any verification process on a Windows development machine will fail, because the type will be detected as `application/octet-stream`.

If I further understand the implementation detail, even if you attempted to do the manual workaround you suggested Adam (setting content-type header, length and explicitly laying out the multi-part boundaries) the FFI will still do the mime byte check and fail with the same issue. 

Am I right in assuming that @rholshausen ?

rholshausen: Yes, that is correct

adamrodger: Worth noting that I don't think the current behaviour conforms to [the spec](https://github.com/pact-foundation/pact-specification/tree/version-3#content-type-of-the-body), which says

> If there is a Content-Type header, the value of the header determines the content type.

That means I'd expect to be always be able to control the content type by passing the appropriate header. In the case of multi-part then obviously the content type header applies to each part separately.

It also doesn't conform to this part:

> default to either application/json (as V1) or text/plain.

Because the default is `application/octet-stream` instead of `text/plain`, and the spec notes that the magic byte test (although the spec incorrectly calls it 'magic number test') is optional.

I think the default should be that the user has to specify a content type, with an opt-in for auto-detection. That way we can document that it may work differently on different OSs and the user can conciously make that choice. If it doesn't work, e.g. because they use different OS or the OS fails auto-detection, then they've always got the option to specify manually.

Inksprout: That section of the spec is referring to the Content-Type of the body of a request, however the issue here is with the Content-Type of the file being uploaded. Requests with a file as part of them include both content types. In the case of using the pactffi_with_multipart_file method from the ffi, the detection of the body Content-Type should work according to the spec. The Content-type of the file being uploaded is what the pactffi_with_multipart_file method helps match. This is the content type that is not able to be detected correctly when using Windows. Hope that helps clear up the issue.

annalauraa:
Any estimate on when this will be completed?

mefellows:
No, the issue is upstream here: https://github.com/pact-foundation/pact-reference/issues/171.

If you are willing to help we'd appreciate it.

YOU54F:
Will need to split out these changes but just wanted to say I revisited this, with some updates to pact-ref which are proposed here for consistent cross platform payload matching - https://github.com/pact-foundation/pact-reference/pull/429

continuation of this PR https://github.com/YOU54F/pact-net/pull/1

also adds in linux arm64 + musl arm64/x64 targets and slimmed binaries and tests them with docker (exception for arm64, which is tested locally as dotnet doesn't support running under qemu), so I need to pull those out. Just wanted to show that we are getting to a better place and hopefully closer to getting this merged.

Thanks @Inksprout for starting it off! Hope you are doing well. 

Thanks for the review so far and concerns raised @adamrodger

adamrodger:
I think this PR should be closed and an RFC raised with a proposed new API so we can discuss. Nobody likes pushing up a PR only to be told the API won't work or something, so let's get agreement on that first. 

I think I started outlining a potential API in the feature request issue for this, so we can carry on the discussion there. The important thing is that there needs to be some kind of fluent builder to allow you to add multiple parts with independent content types and that needs to fit into the existing typestate pattern to prevent you from then being able to set any other body type.

YOU54F:
>I think this PR should be closed and an RFC raised with a proposed new API so we can discuss.

Yeah that is fine with me, I'm not attached to this API in particular, I just used it as a test bed to test out the content type matching in order to get consistent behaviour.

It would be nice to thank the original contributor for their attempt however, when it is closed.

> Nobody likes pushing up a PR only to be told the API won't work or something, so let's get agreement on that first.

I don't think that is a universal statement, I don't mind at all

YOU54F:
Thanks for trying @Inksprout

PR #450, opened 2023-04-06
feat: WithQuery overload with matcher parameter

Manny651:
Add a WithQuery overload `WithQuery(string key, IMatcher matcher)`, similar to `WithHeader(string key, IMatcher matcher)`


adamrodger:
Thanks!

PR #453, opened 2023-04-12
feat: Support v4 specification for consumer tests

adamrodger:
Adds v4 specification support for consumer tests with the same API as v3 sync and async pacts

Requires https://github.com/pact-foundation/pact-reference/issues/264 to be fixed before it can be merged, otherwise CI fails due to the keys regenerating every time.


mefellows:
Nice!

I'll see if I can take a peek at this later today.

One consideration I wanted to flag, is that V4 introduces the ability to add Plugins on interactions. When I created the API for Pact JS (currently in beta), I used a similar type-state model, and the API was definitely influenced by the presence or absence of plugins on an interaction. Or put another way, you may want to consider if adding plugins later on will be a breaking change.

If it helps, here is the HTTP typescript interface: https://github.com/pact-foundation/pact-js/blob/master/src/v4/http/types.ts

Example V4 HTTP interaction with plugins: https://github.com/pact-foundation/pact-js/blob/master/examples/v4/plugins/test/matt.consumer.spec.ts#L27

adamrodger:
Doesn't look like it would be a breaking change I don't think.

It's probably dependant on the type of plugin, but for example if there was a plugin to do non-JSON but still via HTTP (like a proper multi-part plugin) then those would be extensions to `IRequestBuilder` or something so that there's new options available after `UponReceiving`.

Anyway, yeah, not planning on adding any plugins as part of this, but good shout on making sure about that. This is just exactly the same API as V3 so if anything is a breaking change then it would've been anyway I guess.

mefellows:
> It's probably dependant on the type of plugin, but for example if there was a plugin to do non-JSON but still via HTTP (like a proper multi-part plugin) then those would be extensions to IRequestBuilder or something so that there's new options available after UponReceiving.

Yes, you can definitely add plugins to HTTP tests (e.g. to support a new content type - such as [protobufs](https://github.com/pactflow/pact-protobuf-plugin) or as you say, a multi-part body if a plugin was to be built). You need to call two FFI methods:

```
pactffi_using_plugin
// and then (can be set for the request/response body)
pactffi_interaction_contents
```

These can also be used on synchronous (a [new interaction type](https://github.com/pact-foundation/pact-specification/tree/version-4#synchronousmessages) added in V4), asynchronous and HTTP interactions.

mefellows:
🙌

PR #469, opened 2023-07-16
fix(#468): Allow non-ASCII chars on most user input

### LINKED ISSUE #468: Using signs outside ASCII in response body
I'm testing pact-net and I spotted weird behavior when using signs outside ASCII table. I was using my native letters from my country Poland (like ł,ó,ń) - they are UTF-8 compatible but when I'm using them in WithJsonBody mock that is used under the cover is returning empty body. I cannot find in documentation is this by design or is it bug. I suspect some but when interop with native library is used.

adamrodger:
Instead of trying to marshal strings, which don't marshal nicely over the FFI boundary because C# uses UTF-16 but Rust wants UTF-8, instead explicitly convert strings to a UTF-8 `byte[]` and marshal those.

Some places don't need to allow non-ASCII, such as the scheme in URLs, whereas others are very tricky, such as consumer filters. This would change the API to `byte[][]` and those can't be marshalled, so some parts still support non-ASCII for now. If that's a problem in the future then some custom marshalling could be implemented, but currently that seems overkill.

Closes #468

### DIFF
--- a/tests/PactNet.Tests/PactExtensionsTests.cs
+++ b/tests/PactNet.Tests/PactExtensionsTests.cs
@@ -71,6 +71,7 @@ namespace PactNet.Tests
             File.Delete("PactExtensionsTests-Consumer-V3-PactExtensionsTests-Provider.json");
             File.Delete("PactExtensionsTests-Consumer-V4-PactExtensionsTests-Provider.json");
             File.Delete("PactExtensionsTests-Combined-V4-PactExtensionsTests-Provider.json");
+            File.Delete("PactExtensioñsTests-Combined-V4-PactExtensioñsTests-Provider.json");
         }
 
         [Fact]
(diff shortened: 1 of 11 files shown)


tests/PactNet.Tests/PactExtensionsTests.cs
```diff
@@ -71,6 +71,7 @@ public PactExtensionsTests(ITestOutputHelper output)
             File.Delete("PactExtensionsTests-Consumer-V3-PactExtensionsTests-Provider.json");
             File.Delete("PactExtensionsTests-Consumer-V4-PactExtensionsTests-Provider.json");
             File.Delete("PactExtensionsTests-Combined-V4-PactExtensionsTests-Provider.json");
+            File.Delete("PactExtensio�sTests-Combined-V4-PactExtensio�sTests-Provider.json");
```
TDavtyan-ST, line 74: ñ - pretty unnoticable in Code Review :)

TDavtyan-ST (approved):

adamrodger:
This is no longer compatible with PactNet 5.x and an alternative fix will be created instead.

Issue #451, opened 2023-04-09
Add CI for Mac on Apple Sillicon


adamrodger:
Depends on GitHub Actions feature https://github.com/github/roadmap/issues/528

Apple Sillicon (i.e. ARM64-based M1 and M2 processors) will only have alpha-level support until we can run CI fully on this OS/arch combination. This means that we ship the necessary binaries to make it work on a best-effort basis, but any bugs specific to that OS/arch may not be fixed until after this issue is resolved.

Alpha support was added in https://github.com/pact-foundation/pact-net/pull/365 and the binaries were included from [release 4.3.0](https://github.com/pact-foundation/pact-net/releases/tag/4.3.0)

YOU54F:
Hey, 110% agree on this, but don't think we should wait until GitHub Actions adds the appropriate hardware as it might be a while.

Cirrus-CI offers the use of M1 mac's for free for public projects

It is pretty nice

https://cirrus-ci.org/guide/quick-start/

and can be hooked up with https://tart.run/ where a user can run the same CI run on their local machine (similar to https://github.com/nektos/act ) as long as they are running MacOS M1/M2 machines

They also offer the use of an ARM64 Linux Machine (as do CircleCI).

Wonder if it worth setting up an account per pact-foundation project (to avoid concurrency issues on the m1 resources)

As a non-scalable aside - I've managed to setup ephemeral runners on the mac m1 machine for linux/windows/mac github action runners, which at least allow for manual testing / validation of arm64 platforms.

YOU54F:
>I'm not really looking to migrate CI off GitHub Actions currently because the support here is really good. The only problem being ARM-based Mac support.

Not suggesting we migrate everything, but we can add support for an m1 runner via cirrus, or another means for free, with a very easy way to test locally setup locally. Seems like a no brainer

>There was talk about the Pact Foundation paying for a self hosted ARM runner, but not sure where that's got to.

There is money in the pact-foundation kitty [open collective](https://opencollective.com/pact-foundation), however paid for m1 runners are prohibitively expensive from a cursory glance for my own personal needs (read: happy to use SaaS products, don't want to pay _any_ money 😅 )

With a community hat on, we would need someone to invest time in setting up the infra and ascertaining the usage/costs - we don't want a big bill we cant pay, its a fair point, and it's worth bringing up as we get together as a maintainer group to decide what is the best way forward.

[...]

PR #487, opened 2024-02-02
Add mac m1 runner to ci matrix (#451)

TDavtyan-ST:
As github recently introduced support for mac m1/m2 runners (https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/), giving it a try

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
         os:
           - windows-latest
           - ubuntu-latest
-          - macos-latest
+          - macos-12
           - macos-14
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
           - windows-latest
           - ubuntu-latest
           - macos-latest
+          - macos-14
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
         run: dotnet build --no-restore
 
       - name: Test
-        run: dotnet test --no-build --verbosity normal -- RunConfiguration.TargetPlatform=x64
+        run: dotnet test --no-build --verbosity normal
 
       - name: Pack
         if: matrix.os == 'windows-latest'


.github/workflows/ci.yml
```diff
@@ -17,6 +17,7 @@ jobs:
           - windows-latest
           - ubuntu-latest
           - macos-latest
+          - macos-14
```
adamrodger, line 20: What's the difference between this and the latest version above? Are both needed?

.github/workflows/ci.yml
```diff
@@ -36,7 +37,7 @@ jobs:
         run: dotnet build --no-restore
 
       - name: Test
-        run: dotnet test --no-build --verbosity normal -- RunConfiguration.TargetPlatform=x64
```
adamrodger, line 39: Isn't this removing the Mac x86-64 CI? We still need both

.github/workflows/ci.yml
TDavtyan-ST: well, according to https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/
macos-14 is running on m1 machines and macos-latest is now targeting macos12 and is running on x64 architecture

.github/workflows/ci.yml
TDavtyan-ST: as this step is executed for all matrix os's, it will get executed for macos-latest in x64 and for macos-14 in arm64

.github/workflows/ci.yml
YOU54F: `macos-latest` will become `macos-14` at least from my interpretation of the blog post

I believe we should pin to `macos-12` to ensure we retain x86_64 builds, when `macos-latest` switches

> The macOS 12 runner image will remain latest until migration of the latest YAML workflow label to macOS 14 in Q2 FY24 (April – June 2024). While macOS 13 is now generally available under the macos-13 label, this image will not be migrated to latest. Following this announcement, macOS 11 runner image will begin deprecation immediately with retirement expected to complete by June 2024.

The runner images show readme's for ARM/AMD versions of macos-14 but there doesn't seem to be a way to specify an arch (that I've seen), so if intel mac's get phased out on gh actions, we may have to rely on rosetta

We've done that against the other two repos currently using these runners

- https://github.com/pact-foundation/pact-php/pull/457
- https://github.com/pact-foundation/pact-js-core/pull/478

.github/workflows/ci.yml
YOU54F: yep thats how I read that, it will execute for both steps and use the systems platform+arch, the latter of which will differ in the case of the two distinct macos images which are arch specific

YOU54F (approved):

adamrodger:
I think we need to make it more explicit which job is using which arch. Just from looking at the CI definition it's unclear that 12 just happens to be x86 and 14 is ARM, plus those could change in future.

adamrodger:
Added in `5.0.0-beta.2` with explicit arch support using GitHub Actions matrix feature

Issue #496, opened 2024-04-22
RFC: Add linux-musl-x64 Runtime Support


botondbotos:
**Previous issues**
[Add `linux-musl-x64` Runtime Support](https://github.com/pact-foundation/pact-net/issues/374)

**Is your feature request related to a problem? Please describe.**
According to [Compatibility / Operating System](https://github.com/pact-foundation/pact-net?tab=readme-ov-file#operating-system) page *Linux (musl)* is not supported, however starting with [Pact FFI Library 0.4.17](https://github.com/pact-foundation/pact-reference/releases/tag/libpact_ffi-v0.4.17) the `libpact_ffi-linux-aarch64-musl.so` and `libpact_ffi-linux-x86_64-musl.so` shared libraries are part of releases.

**Describe the solution you'd like**
Update pact-net to use latest [Pact FFI Library 0.4.19](https://github.com/pact-foundation/pact-reference/releases/tag/libpact_ffi-v0.4.19) which has the `libpact_ffi` shared libraries and extend the NuGet package with additional runtime targets.

**Breaking Changes**
No

**Potential Downsides/Caveats**
n/a

**Describe alternatives you've considered**
n/a

**Additional context**
After replacing the `runtimes/linux-x64/native/libpact_ffi.so` file added by [PactNet 5.0.0-beta.2](https://www.nuget.org/packages/PactNet/5.0.0-beta.2) package with [libpact_ffi-linux-x86_64-musl.so](https://github.com/pact-foundation/pact-reference/releases/download/libpact_ffi-v0.4.19/libpact_ffi-linux-x86_64-musl.so.gz) I was able to successfully run pact-net tests inside Docker using `mcr.microsoft.com/dotnet/sdk:6.0.421-alpine3.19-amd64`

adamrodger:
Currently there is no reliable way to detect libc vs musl during a .Net restore operation and therefore no reliable way to ensure the correct libpact binary is restored. That makes this issue currently infeasible until a reliable way of doing that detection is found.

botondbotos:
Could you please help me understand why  it is preferred to restore the correct lib_ffi package over picking the correct one at runtime? 
I have a PoC implemented, which is addressing the issue of musl support: https://github.com/pact-foundation/pact-net/pull/497#issuecomment-2087340260
One of the drawbacks is that all native packages need to be deployed, but the runtime can pick the correct one.

adamrodger:
[...]

PR #497, opened 2024-04-22
Add linux-musl-x64 runtime support

botondbotos:
Possible solution for https://github.com/pact-foundation/pact-net/issues/496

### DIFF
--- a/src/PactNet/PactNet.csproj
+++ b/src/PactNet/PactNet.csproj
@@ -36,6 +36,13 @@
       <CopyToOutputDirectory Condition="'$(IsLinux)'">PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
     </Content>
+    <Content Include="$(MSBuildProjectDirectory)\..\..\build\linux\x86_64-musl\libpact_ffi.so">
+      <Link>libpact_ffi.so</Link>
+      <PackagePath>runtimes/linux-musl-x64/native</PackagePath>
+      <Pack>true</Pack>
+      <CopyToOutputDirectory Condition="'$(IsLinux)'">PreserveNewest</CopyToOutputDirectory>
+      <Visible>false</Visible>
+    </Content>
     <Content Include="$(MSBuildProjectDirectory)\..\..\build\osx\x86_64\libpact_ffi.dylib">
       <Link>libpact_ffi.dylib</Link>
       <PackagePath>runtimes/osx-x64/native</PackagePath>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ Before raising any issues, please make as much effort as you can to rule out iss
 For example, if you are using a self-hosted Pact Broker instance and PactNet is failing to connect, please ensure things like the
 authentication token and SSL certificate are valid.
 
-If you are sure that the issye is with PactNet then please raise an issue, including as many of the following details as you can:
+If you are sure that the issue is with PactNet then please raise an issue, including as many of the following details as you can:
 
 - PactNet version
 - Your operating system and version
@@ -59,7 +59,7 @@ version by executing the script in Bash (or Git Bash on Windows):
 build/download-native-libs.sh
 ```
 
-Alternatively you can download a particular FFI version from the [pact-referece] releases or build your own version
+Alternatively you can download a particular FFI version from the [pact-reference] releases or build your own version
 locally, and then copy the artifacts into the folders:
 
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,18 +63,19 @@ Alternatively you can download a particular FFI version from the [pact-reference
 locally, and then copy the artifacts into the folders:
 
 ```
-build/
-    linux/
-        x86_64/
-            libpact_ffi.so
-    osx/
-        aarch64-apple-darwin/
-            libpact_ffi.dylib
-        x86_64/
-            libpact_ffi.dylib
-    windows/
-        x86_64/
-            pact_ffi.dll
+├── linux
+│   ├── x86_64
+│   │   └── libpact_ffi.so
+│   └── x86_64-musl
+│       └── libpact_ffi.so
+├── osx
+│   ├── aarch64-apple-darwin
+│   │   └── libpact_ffi.dylib
+│   └── x86_64
+│       └── libpact_ffi.dylib
+└── windows
+    └── x86_64
+        └── pact_ffi.dll
 ```
 
 After the native libraries are in the expected places then the solution can be built in your IDE or on the command
--- a/build/download-native-libs.sh
+++ b/build/download-native-libs.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-FFI_VERSION="0.4.16"
+FFI_VERSION="0.4.19"
 FFI_BASE_URL="https://github.com/pact-foundation/pact-reference/releases/download/libpact_ffi-v$FFI_VERSION"
 
 GREEN="\e[32m"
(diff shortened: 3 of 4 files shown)


src/PactNet/PactNet.csproj
```diff
@@ -36,6 +36,13 @@
       <CopyToOutputDirectory Condition="'$(IsLinux)'">PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
     </Content>
+    <Content Include="$(MSBuildProjectDirectory)\..\..\build\linux\x86_64-musl\libpact_ffi.so">
+      <Link>libpact_ffi.so</Link>
+      <PackagePath>runtimes/linux-musl-x64/native</PackagePath>
+      <Pack>true</Pack>
+      <CopyToOutputDirectory Condition="'$(IsLinux)'">PreserveNewest</CopyToOutputDirectory>
```
adamrodger, line 43: This needs an additional condition to only unpack on musl targets, which last time I tried to do this wasn't possible.

The libc and musl versions are both just called libpact_ffi.so and so their names conflict and it's unclear which one you actually get.

If you get the musl version where previously you get the libc version then this is a breaking change.

It's the same reason why the Mac version has two conditions below, so you don't end up with the ARM version on x86 and vice versa.

adamrodger:
The CI additionally would need updating to test the new target

YOU54F:
Nice,

I had a go at this myself as part of a mission to bring alpine/musl support to the Pact ecosystem - see linked [issue](https://github.com/pact-foundation/devrel/issues/30)

https://github.com/pact-foundation/pact-net/compare/master...YOU54F:pact-net:feat/musl

added a gh test in the branch, tried to test arm64 under qemu but no 🎲 

https://github.com/pact-foundation/pact-net/compare/master...YOU54F:pact-net:feat/trimmed_binaries

There still doesn't seem to be a nice way to detect musl via .NET

I used a file system lookup for the musl shared lib, which varies on the arch of the target, and I've only tested against alpine based images. 

From looking at a few different languages, most tend to `ldd <executable>` like `/bin/sh` and grep for musl, unless its a compiled language and compilation is set by flags (like rust). I couldn't find a way in the .NET sln to run a command, so went for a musl shared lib look up instead.

For rust based executables, (not the pact_ffi shared libs), we are beginning to roll out the building of them statically with musl, rather than glibc, so we can support a single executable for all linux users.

Unfortunatly, we cannot do that with the ffi libs in languages that read on the shared, not static libs (PHP, .NET, Ruby)

adamrodger:
Yep that was the same problem I had when I originally raised a PR for this and had to abandon it @YOU54F.

There doesn't seem to be a reliable way to determine libc vs musl at NuGet restore time to make sure the correct shared library is unpacked, and if you get the wrong one it just won't work with no real workaround. That would be a disaster for existing users, hence why we need extensive CI for that situation.

It's got to be totally reliable and it can't break existing libc distros, and that's where the problem lies.

adamrodger:
It may be more pragmatic on Alpine to install something like https://github.com/Stantheman/gcompat to provide glibc compatibility so that the existing shared library will work, although I've not tested that.

botondbotos:
> It may be more pragmatic on Alpine to install something like https://github.com/Stantheman/gcompat to provide glibc compatibility so that the existing shared library will work, although I've not tested that.

Unfortunately, `gcompat` doesn't seem to work: https://github.com/botondbotos/pact-net-glibc-vs-musl/blob/main/Dockerfile.gcompat
I managed to make it run by patching `libpact_ffi.so` with https://github.com/pact-foundation/pact-reference/releases/download/libpact_ffi-v0.4.19/libpact_ffi-linux-x86_64-musl.so.gz

botondbotos:
Refactored the project to use runtime specific native packages as shown at https://github.com/Mizux/dotnet-native. 
Managed to run the tests under `samples/OrdersApi`:

```
./pack.sh
docker build -f Dockerfile.alpine  -t pact-net-alpine .
docker build -f Dockerfile.debian -t pact-net-debian .
```

The `runtimes` folder seems to contain all native libraries.

**On Alpine**

```
docker run --rm  pact-net-alpine tree samples/OrdersApi/Consumer.Tests/bin/Debug/net8.0/runtimes
```
Output:
```
samples/OrdersApi/Consumer.Tests/bin/Debug/net8.0/runtimes
├── linux-musl-x64
│   └── native
│       └── libpact_ffi.so
├── linux-x64
│   └── native
│       └── libpact_ffi.so
├── osx-arm64
│   └── native
│       └── libpact_ffi.dylib
├── osx-x64
│   └── native
│       └── libpact_ffi.dylib
├── win
│   └── lib
│       ├── net6.0
│       │   ├── System.Diagnostics.EventLog.Messages.dll
│       │   └── System.Diagnostics.EventLog.dll
│       └── netstandard2.0
│           └── System.Security.Cryptography.ProtectedData.dll
└── win-x64
    └── native
        └── pact_ffi.dll
```

**On Debian**

```
docker run --rm  pact-net-debian tree samples/OrdersApi/Consumer.Tests/bin/Debug/net8.0/runtimes
```

Output:

```
samples/OrdersApi/Consumer.Tests/bin/Debug/net8.0/runtimes
|-- linux-musl-x64
|   `-- native
|       `-- libpact_ffi.so
|-- linux-x64
|   `-- native
|       `-- libpact_ffi.so
|-- osx-arm64
|   `-- native
|       `-- libpact_ffi.dylib
|-- osx-x64
|   `-- native
|       `-- libpact_ffi.dylib
|-- win
|   `-- lib
|       |-- net6.0
|       |   |-- System.Diagnostics.EventLog.Messages.dll
|       |   `-- System.Diagnostics.EventLog.dll
|       `-- netstandard2.0
|           `-- System.Security.Cryptography.ProtectedData.dll
`-- win-x64
    `-- native
        `-- pact_ffi.dll
```

Is it worth pursuing this approach?

adamrodger:
PactNet used to have architecture specific packages but a key design goal of the major changes in 4.x was to move away from this approach.

In practice it's very awkward to use for consumers - for example if you run the tests locally on Windows/Mac but your CI runs in Linux (as is very common) then you've got to reference multiple native packages and add conditions to your dependencies, which is a bad devex.

I don't see the cost of reintroducing all the complexity both within PactNet and especially for our users being worth adding musl support. At the end of the day this is a test library, so it's much less of a burden to run your tests in a glibc env instead, until such a time that detecting running on musl is more robust and can use the existing packaging mechanism.

It may be worth you creating an issue on the .Net team to add a feature to make detecting musl easier. It's definitely not an easy task to do really robustly though. Like even just detecting certain heuristics is hard because of things like cross compilation envs.

botondbotos:
>In practice it's very awkward to use for consumers - for example if you run the tests locally on Windows/Mac but your CI runs in Linux (as is very common) then you've got to reference multiple native packages and add conditions to your dependencies, which is a bad devex.

There's only need to reference a single package, `PactNet`: https://github.com/pact-foundation/pact-net/pull/497/files#diff-bc4f87702468a93737635b96e0dc7fc5ec46a8fab44ebc9f249a17091de6fd60R16. 
The `PactNet` package depends on all the other packages containing the native pact ffi libraries. The packages containing the native libraries should never be referenced directly. 
This packaging model helps generate the correct `deps.json` so that the correct assembly can be picked up by the runtime. This is demonstrated by the [Dockerfile.alpine](https://github.com/pact-foundation/pact-net/pull/497/files#diff-f865864730d7062c04cc13b1b85ba03ed5dbef3fa02fcb50b87d9117d0af1f4e) and [Dockerfile.debian](https://github.com/pact-foundation/pact-net/pull/497/files#diff-49a95235edddaf12f383a8f467d25b6ae8422204561d88fdaadfc9c1c6a9afe2) containers.

>... At the end of the day this is a test library, so it's much less of a burden to run your tests in a glibc env instead ...

Some teams prefer to run tests on the exact same platform with exactly the same dependencies as what gets deployed to production. Exceptions could be made but that adds extra complexity to the CI pipeline.

PR #506, opened 2024-06-10
Fix/numeric provider state param

### LINKED ISSUE #449: Provider state parameters are serialised as numbers, causes errors in provider state middleware when using System.Text.Json
If the value of a provider state parameter is a number string, it is serialised in the pact file as a number. This is an issue if the provider state middleware in the provider tests uses System.Text.Json to deserialise a `ProviderState` value as System.Text.Json does not deserialise non-string values into string properties. https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-6-0#non-string-values-for-string-properties.

For example:

```csharp
using System.Net;
using Newtonsoft.Json.Serialization;
using Newtonsoft.Json;
using PactNet;

var config = new PactConfig
{
    PactDir = Path.Join("..", "..", "..", "pacts"),
    DefaultJsonSettings = new JsonSerializerSettings
    {
        ContractResolver = new CamelCasePropertyNamesContractResolver(),

    }
};

var pact = Pact.V3("consumer", "provider", config).WithHttpInteractions();

pact.UponReceiving("A valid request")
    .Given("state with parameter", new Dictionary<string, string>() {["id"] = "10", ["name"] = "Fred"})
    .WithRequest(HttpMethod.Get, "/api/endpoint")
    .WillRespond()
    .WithStatus(HttpStatusCode.OK);

await pact.VerifyAsync(async ctx =>
{
    using var apiClient = new ApiClient(ctx.MockServerUri);

    await apiClient.TestRequestAsync();
});

public class ApiClient: IDisposable
{
    private readonly HttpClient _httpClient;

    public ApiClient(Uri baseUri)
    {
        _httpClient = new HttpClient { BaseAddress = baseUri };
    }

    public async Task TestRequestAsync()
    {
        await _httpClient.GetAsync("api/endpoint");
    }

    public void Dispose()
    {
        _httpClient.Dispose();
    }
}
```

generates this pact. Note that parameter "id" is serialised as `10` not `"10"`.

```json
{
  "consumer": {
    "name": "consumer"
  },
  "interactions": [
    {
      "description": "A valid request",
      "providerStates": [
        {
          "name": "state with parameter",
          "params": {
            "id": 10,
            "name": "Fred"
          }
        }
      ],
      "request": {
        "method": "GET",
        "path": "/api/endpoint"
      },
      "response": {
        "status": 200
      }
    }
  ],
  "metadata": {
    "pactRust": {
      "ffi": "0.4.0",
      "models": "1.0.4"
    },
    "pactSpecification": {
      "version": "3.0.0"
    }
  },
  "provider": {
    "name": "provider"
  }
}
```

YOU54F:
fixes #449 

- Added examples of generating provider states with number or string values
- Serialise value into json, prior to sending over the wire
  - As per recommendation in https://github.com/pact-foundation/pact-reference/issues/263#issuecomment-1510555100
  - As per pact-js https://github.com/pact-foundation/pact-js/blob/54087fa06e485f1c7041a0a6f7f14e1cc177e09f/src/v3/pact.ts#L100-L102

relates to

- https://github.com/pact-foundation/pact-reference/issues/298
- https://github.com/pact-foundation/pact-reference/issues/263


src/PactNet/Drivers/HttpInteractionDriver.cs
```diff
@@ -37,8 +37,11 @@ public void Given(string description)
         /// <param name="name">Parameter name</param>
         /// <param name="value">Parameter value</param>
         /// <returns>Success</returns>
-        public void GivenWithParam(string description, string name, string value)
-            => NativeInterop.GivenWithParam(this.interaction, description, name, value).CheckInteropSuccess();
+        public void GivenWithParam(string description, string name, object value)
+            {
+                var jsonValue = System.Text.Json.JsonSerializer.Serialize(value);
```
adamrodger, line 42: style: Don't use var when it's not obvious what the type is 

style: Don't use fully qualified references inline, use an import instead

src/PactNet/Drivers/MessageInteractionDriver.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Runtime.InteropServices;
 using PactNet.Interop;
-
+using System.Text.Json.Serialization;
```
adamrodger, line 4: style: Blank line after import statements

adamrodger (changes requested):
If the API is to open up to allow absolutely anything be used as a provider state parameter and change the public API then I'd want to see more significant testing of "weird" values.

So there are obvious things like what does `double.NaN` do? What about null or other nullable value types? What about complex objects? What about random classes being passed in? What about string-like types such as `StringBuilder` and `Span<char>`? What about arrays or other collection types? Stuff like that.

We also need a lot of round trip testing, to make sure that the provider tests can actually deserialise these things. I know we expect users to handle provider states themselves but we at least need to make sure that `System.Text.Json` can round trip these things given that's the main serialiser we support.

Using `object` in a public API creates a lot of opportunity for that to all go wrong so I think we need much more comprehensive testing when we're serialising potentially anything.

src/PactNet.Abstractions/IMessageBuilder.cs
```diff
@@ -21,7 +21,7 @@ public interface IMessageBuilderV3
         /// <param name="providerState">Provider state description</param>
         /// <param name="parameters">Provider state parameters</param>
         /// <returns>Fluent builder</returns>
-        IMessageBuilderV3 Given(string providerState, IDictionary<string, string> parameters);
+        IMessageBuilderV3 Given(string providerState, IDictionary<string, object> parameters);
```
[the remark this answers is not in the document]
scrocquesel-ml150: Could it be a `dynamic` ? It would be aligned with `WithContent`.
Could it be planned for the next major version ?

Issue #505, opened 2024-06-06
RFC: Distinguish technical failures when verifying pacts


flakey-bit:
As provider, when we verify ourselves against published consumer Pacts (`IPactVerifierSource::Verify`), a `PactFailureException` can be thrown.

This exception doesn't distinguish between
1. The provider simply not honouring one or more contracts &
2. _Technical_ failures (i.e. the server returning 500 responses) while evaluating the contracts

This is an **important distinction** - we'd like to fail our build if there are _technical_ failures while self-evaluating against contracts, however if we simply don't honour a given contract then that's not reason to fail the build (the contract compatibilty will be published to the broker and we'll rely on `can-i-deploy` at deployment time).

Probably the simplest thing would be to introduce fields to `PactFailureException` providing more information on what happened (which contracts failed, what the status codes were etc). That shouldn't be a breaking change 🤞 
 
No obvious downsides, other than the plumbing needed to pass the data in. [As mentioned on the Slack](https://pact-foundation.slack.com/archives/C9UTHV2AD/p1717643316417299?thread_ts=1717606895.997469&cid=C9UTHV2AD), we might want something in core so that other implementations (not just dotnet) can benefit. 
 
**Describe alternatives you've considered**
As a workaround, we've added a custom logging sink to our server when it evaluates the contracts. If any errors are logged then we fail the unit test (we're following the recommended approach whereby contract compatibility is evaluated in a unit test during the build).

**Additional context**
This came about because our provider codebase mocks/stubs various services when evaluating itself against Pact contracts. We were missing some stub setup calls, resulting in errors being logged / 500 responses.

adamrodger:
Yeah I think having some extra info would be useful as to why verification failed. Obviously it would still throw an exception but you could choose to catch and ignore certain exceptions in this specific use case.

As you say, this would need changes to the core library instead of PactNet specifically at first, so it's probably best raising an issue there and linking here so it can be implemented once it's supported in FFI.

flakey-bit:
@adamrodger is https://github.com/pact-foundation/pact-reference/issues/132 relevant?

adamrodger:
No, that was an issue whereby FFI always logged directly to stdout so you couldn't report the failure logs in test output, and that was sorted quite a while ago.

flakey-bit:
Having a look at the FFI code a bit (apologies, not a rust developer) 

[...]

PR #508, opened 2024-06-23
[GH-505]: Include VerificationExecutionResult when verification fails

flakey-bit:
As per https://github.com/pact-foundation/pact-net/issues/505 we want to be able to distinguish tehcnical failures from simply not honouring the contract.

This PR throws a more specific exception which allows the caller to determine what has happened.

Merging into the 4.x release branch rather than `master` - if not, would need to revisit JSON serialization side of things


adamrodger:
So I think for now we should just throw the more specific exception rather than try to deserialise all the verification results. They'll also be printed to the console, and it's a big UX change to alter that. 

This would also need to target 5.x because it's a breaking API change now that we're throwing a different exception type (albeit a sub type).

adamrodger:
Superceded by https://github.com/pact-foundation/pact-net/commit/0199f010c82551c018d3b56bc903aefaf6212750

PR #518, opened 2024-09-07
feat: add array contains matcher

MaxCampman:
# What?
* Add array-contains matcher (e.g. [JS version docs](https://docs.pact.io/implementation_guides/javascript/docs/matching#array-contains-matcher)).
* I was able to use this matcher via a workaround but thought I should add it here to make it available to others.
* This was added to the pact-reference project here: https://github.com/pact-foundation/pact-reference/pull/459
* In order to incorporate this change, I have had to update the pact-ffi version to 0.4.23 (from 0.4.16)
* This required some small tweaks to the integration tests and some tweaking of the script used to fetch the libraries (there were a few name changes for libraries).
* Actual change is implemented as:
  * Addition of: `src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs`
  * Addition to: `src/PactNet.Abstractions/Matchers/Match.cs`
  * Addition to: `src/PactNet.Abstractions/Matchers/MatcherConverter.cs`

# NOTE:
* I unfortunately do not have access to an Apple computer at the moment.
* This likely means that the build script needs some further tweaking for the OSX-specific script (`if [[ "$OSTYPE" == "darwin"* ]]; then`)


src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
```diff
@@ -0,0 +1,17 @@
+using System.Text.Json.Serialization;
+
+namespace PactNet.Matchers;
```
adamrodger, line 3: style: The project uses namespace blocks instead of file scoped namespaces

src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
```diff
@@ -0,0 +1,17 @@
...
+
+public class ArrayContainsMatcher : IMatcher
```
adamrodger, line 5: issue: The public API should be documented

src/PactNet.Abstractions/Matchers/Match.cs
```diff
@@ -168,5 +168,10 @@ public static IMatcher Include(string example)
         {
             return new IncludeMatcher(example);
         }
+
+        public static IMatcher ArrayContains(dynamic[] variations)
```
adamrodger, line 172: issue: The public API should be documented

tests/PactNet.Abstractions.Tests/Matchers/ArrayContainsMatcherTests.cs
```diff
@@ -0,0 +1,29 @@
...
+                "Thing2",
+            };
+
+            var matcher = new ArrayContainsMatcher(example);
+
+            // Act
+            var actual = JsonSerializer.Serialize(matcher);
+
+            // Assert
+            actual.Should().Be(@"{""pact:matcher:type"":""array-contains"",""variants"":[""Thing1"",""Thing2""]}");
```
adamrodger, line 26: suggestion: You can use new string literals to format the JSON nicely here

adamrodger (changes requested):
The FFI change should probably be split into a separate preceding change just because of the extra complexity around the breaking changes in the FFI packaging. I could see that causing issues so a dedicated change would be better.

In terms of the example, the pact spec describes this matcher as checking that all variants are contained within a collection, but this example doesn't really reflect that given the collection is only of one type and thus only one variant. The eachLike matcher is better in that scenario. I think the example should somehow contain different types (and if it's too difficult to clearly do that then perhaps the matcher doesn't make as much sense in a typed language Vs something like JavaScript).

I've not reviewed the integration tests as I'm on my phone and they're quite big, so there may be further comments there later.

Also check any line endings settings in your editor because a few places have weird extra blank lines and things, plus obviously the sln file shows a huge diff when there probably isn't one.

samples/OrdersApi/Consumer.Tests/pacts/Fulfilment API-Orders API.json
```diff
@@ -71,6 +71,126 @@
...
+                  "match": "arrayContains",
+                  "variants": [
+                    {
+                      "index": 0,
+                      "rules": {
+                        "$.date": {
+                          "combine": "AND",
+                          "matchers": [
+                            {
+                              "match": "type"
```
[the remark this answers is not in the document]
MaxCampman: This is from this line: 

```csharp
Date = Match.Type(expected1.Date.ToString("O"))
```

So is indeed meant to be type

samples/OrdersApi/Consumer.Tests/pacts/Fulfilment API-Orders API.json
```diff
@@ -88,6 +208,126 @@
       },
       "type": "Synchronous/HTTP"
     },
+    {
+      "description": "a request for multiple orders by id",
```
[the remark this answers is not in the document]
MaxCampman: This is my fault, can see the description here is different to the one above.
Will amend

src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
MaxCampman: Noted, will amend

src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
MaxCampman: Noted, will amend

src/PactNet.Abstractions/Matchers/Match.cs
MaxCampman: Noted, will amend

tests/PactNet.Abstractions.Tests/Matchers/ArrayContainsMatcherTests.cs
MaxCampman: Thanks, quite the oversight from me

MaxCampman (commented):
Thanks for taking the time to go through this, especially on a weekend.

I will open a new MR to just update the FFI, and will come back to this one to make the requested changes

adamrodger:
@MaxCampman if you rebase onto latest then the FFI upgrade should already be there 👍

samples/OrdersApi/Consumer.Tests/pacts/Fulfilment API-Orders API.json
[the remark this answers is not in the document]
MaxCampman: Fixed

samples/OrdersApi/Consumer.Tests/OrdersClientTests.cs
```diff
@@ -84,6 +84,58 @@ await this.pact.VerifyAsync(async ctx =>
...
+            this.pact
+                .UponReceiving("a request for multiple orders by id")
+                .Given("orders with ids {ids} exist", new Dictionary<string, string> { ["ids"] = "1,2" })
+                .WithRequest(HttpMethod.Get, "/api/orders/many/1,2")
+                .WithHeader("Accept", "application/json")
+                .WillRespond()
+                .WithStatus(HttpStatusCode.OK)
+                .WithJsonBody(Match.ArrayContains(new dynamic[]
+                {
+                    new
```
adamrodger, line 102: issue: I think for this example to 'come to life' then we'd really need it to match multiple different possible variants

Perhaps if there were different order types or something so that they could have different fields on them, then the matching rules could reflect those variants. Like a fulfilled order has an extra date on it for when it was completed, or something like that.

samples/OrdersApi/Provider.Tests/ProviderStateMiddleware.cs
```diff
@@ -52,6 +54,15 @@ private async Task EnsureEventExistsAsync(IDictionary<string, object> parameters
             await this.orders.InsertAsync(new OrderDto(id.GetInt32(), OrderStatus.Fulfilling, DateTimeOffset.Now));
         }
 
+        private async Task EnsureEventsExistAsync(IDictionary<string, object> parameters)
```
adamrodger, line 57: docs: Needs doc string like other methods

samples/OrdersApi/Provider/Orders/OrdersController.cs
```diff
@@ -46,6 +47,29 @@ public async Task<IActionResult> GetByIdAsync(int id)
             }
         }
 
+        [HttpGet("many/{ids}", Name = "getMany")]
```
adamrodger, line 50: docs: Needs doc string like other methods

src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
```diff
@@ -0,0 +1,28 @@
+using System.Text.Json.Serialization;
+
+namespace PactNet.Matchers
+{
+    public class ArrayContainsMatcher : IMatcher
```
adamrodger, line 5: docs: Needs doc string like other matchers

src/PactNet.Abstractions/Matchers/Match.cs
adamrodger: This one still needs documenting

tests/PactNet.Abstractions.Tests/Matchers/ArrayContainsMatcherTests.cs
adamrodger: Ah, yes my bad. These have to support .Net Framework also

tests/PactNet.Tests/Matchers/ArrayContainsMatcherTests.cs
```diff
@@ -0,0 +1,252 @@
...
+    {
+        // Arrange
+        this.pactBuilder
+            .UponReceiving("A request for strings")
+            .WithRequest(HttpMethod.Get, "/api/list/string")
+            .WillRespond()
+            .WithStatus(HttpStatusCode.OK)
+            .WithJsonBody(Match.ArrayContains([
+                Match.Regex("Thing 1", "Thing [0-9]+"),
+                Match.Regex("Thing X", "Thing [A-Z]+"),
```
adamrodger, line 60: comment: This is a good test because it shows using different variants, which we should replicate in the example usage

adamrodger (changes requested):

PactNet.sln
```diff
@@ -1,177 +1,177 @@
...
-		{B8C8934E-9581-46D6-A728-BC5EAB9C06DD} = {A117BBC6-60BB-4282-BF10-E616DE0AFAD0}
-		{5E915D66-917B-4730-B31A-C9727C196346} = {6663C12E-9912-40D0-9310-D119D1F6B023}
-		{D8B75E48-6E45-468B-8049-B73823C14CB8} = {6663C12E-9912-40D0-9310-D119D1F6B023}
-		{02E265A1-A7A2-4106-8F6A-5027FDC3FC50} = {CF67D7A1-AE96-420B-9971-65E535B903E8}
-	EndGlobalSection
-	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {C2CBC30C-92D4-4E3A-A5B8-1E5D4E938DFC}
-	EndGlobalSection
-EndGlobal
+﻿
```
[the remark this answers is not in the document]
MaxCampman: Note sure what is causing this one, think it may be my local git settings

src/PactNet.Abstractions/Matchers/Match.cs
MaxCampman: Done

samples/OrdersApi/Provider.Tests/ProviderStateMiddleware.cs
MaxCampman: done

samples/OrdersApi/Provider/Orders/OrdersController.cs
MaxCampman: done

src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
MaxCampman: done

src/PactNet.Abstractions/Matchers/ArrayContainsMatcher.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// The items expected to be in the array.
+        /// </summary>
+        [JsonPropertyName("variants")]
+        public dynamic Value { get; }
+
+        /// <summary>
+        /// Initialises a new instance of the <see cref="ArrayContainsMatcher"/> class.
+        /// </summary>
+        /// <param name="variants"></param>
+        public ArrayContainsMatcher(dynamic[] variants)
```
adamrodger, line 26: question: This passes in an array, but we store the `Value` field as a single `dynamic` - is that correct? I would've thought it would have been a collection also.

suggestion: Use `ICollection` instead of an array for a more ergonomic API

tests/PactNet.Tests/Matchers/ArrayContainsMatcherTests.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        await this.pactBuilder.VerifyAsync(async ctx =>
+        {
+            using var service = Service(ctx.MockServerUri);
+
+            // Act
+            var strings = await service.GetStringList();
+
+            // Assert
+            strings.Should().Contain("Thing 1").And.Contain("Thing 2");
```
adamrodger, line 45: suggestion: Use `Should().Equal(["Thing 1", "Thing 2"]}` to be more explicit that it should contain all of the expected, and only the expected

PR #525, opened 2024-10-28
chore(docs): mark pact-net 5.x as stable

YOU54F:
Should 4.x be marked as deprecated, or is that still supported for a certain period of time, to allow for migration to 4.x and potentially allow for changes/fixes to be released to the 4.x line?

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ Due to using a shared native library instead of C# for the main Pact logic only 
 
 | Version | Status     | [Spec] Compatibility | Install            |
 | ------- | ---------- | -------------------- | ------------------ |
-| 5.x     | Beta       | 2, 3, 4              | See [installation] |
+| 5.x     | Stable     | 2, 3, 4              | See [installation] |
 | 4.x     | Stable     | 2, 3                 | See [installation] |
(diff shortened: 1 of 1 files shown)


adamrodger:
Oh I totally forgot to do that, thanks!

I think 4.x would only receive updates now for a major issue or big security problem. The breaking changes in 5.x are pretty easy to migrate, and it's complicated if the 4.x fix needed an FFI update.

So that's probably a kind of LTS? Unless you can think of a better term. Problem with LTS is we have no defined period of support. Maybe 'maintenance'? I'm struggling to think of a good term.

YOU54F:
I think marking as deprecated will just prompt those to migrate to v5. You’ve created upgrade guides and I’ve updated the workshops and added the contents from the docs folder to the [website](https://docs.pact.io/implementation_guides/net) in this [pr](https://github.com/pact-foundation/docs.pact.io/pull/359/files), so think we are in a good position

adamrodger:
Yeah fine by me. Do you want to update the PR with that change at the same time then?

YOU54F:
> Yeah fine by me. Do you want to update the PR with that change at the same time then?

done now 👍🏾

PR #533, opened 2025-02-18
Raw XML POST not working in V2, V3 and V4. Added test for simple POST interaction with different content types.

vcalatayud:
Added a test to ensure that POST works with different types of content. XML is failing with V2, V3 and V4.


adamrodger:
Sorry, I'm unclear what the purpose of the PR is, could you explain? Are you saying that if this new test is merged then it will fail?

vcalatayud:
Hi :) yes, one of these tests fails. I thought it could be useful to reproduce the issue. And maybe to work on the solution in the same branch and merge it when the test passes.

Issue #459, opened 2023-05-10
Messaging pacts - WithContentAsync - factory is invoked before provider state has been set


DavidJFowler:
When verifying a messaging pact scenario using WithContentAsync, the factory Func is invoked before a POST to /provider-states has completed.

To reproduce, create a dotnet 7 console app. Add a FrameworkReference to Microsoft.AspNetCore.App in the .csproj file:

```xml
<ItemGroup>
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
  </ItemGroup>
```

add the following package refs:

```xml
<ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.11.0" />
    <PackageReference Include="NUnit" Version="3.13.3" />
    <PackageReference Include="NUnit3TestAdapter" Version="4.4.2" />
    <PackageReference Include="PactNet" Version="4.5.0" />
  </ItemGroup>
```

Edit program.cs:

```csharp
using Newtonsoft.Json.Serialization;
using Newtonsoft.Json;
using NUnit.Framework;
using PactNet;
using PactNet.Matchers;
using PactNet.Verifier;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using System.Net;
using Microsoft.AspNetCore.Http;

// create pact
var pactDir = Path.Join("..", "..", "..", "pacts");
var v3 = Pact.V3("Message Consumer", "Message Producer", new PactConfig
{
    PactDir = pactDir,
    DefaultJsonSettings = new JsonSerializerSettings
    {
        ContractResolver = new CamelCasePropertyNamesContractResolver()
    }
});

var messagePact = v3.WithMessageInteractions();

messagePact.ExpectsToReceive("some events")
    .Given("events exist")
    .WithJsonContent(Match.MinType(new { MessageText = "Hello World"}, 1))
    .Verify<ICollection<MyMessage>>(events => Assert.That(events, Is.Not.Empty));

// verify pact

// configure provider states handler
var isProviderStatesCalled = false;
const string pactProviderServiceUri = "http://127.0.0.1:9001";
var builder = WebApplication.CreateBuilder(new WebApplicationOptions());
builder.WebHost.UseUrls(pactProviderServiceUri);
await using var app = builder.Build();
app.MapPost("/provider-states", async context =>
{
    isProviderStatesCalled = true;
    context.Response.StatusCode = (int) HttpStatusCode.OK;
[...]

PR #536, opened 2025-04-01
feat!: Invoke async message scenario factories on request

adamrodger:
We still have to invoke the factory with sync-over-async at request time and this puts a burden on the user not to use this improperly, but if we want to delay invocation of the factory until after provider states have been configured then this is the first step.

This is also obviously a breaking API change and requires a major version bump. `WithAsyncContent` makes way more sense because it's the content factory that's async, not the call to the builder as `WithContentAsync` would imply. This also means the return type changes to void instead of Task.

See #459

We should also investigate making the internal messaging server itself async so that we don't have to do the sync-over-async shenanigans, which could maybe be achieved with `Task.Run` (or similar) instead of using a dedicated thread. We'd probably then also change the signature of `Scenario.Factory` to be `Func<Task<dynamic>>` instead, and the sync versions would wrap in `Task.FromResult`.

We do get really weird behaviour when using `dynamic` inside `Func<>` though, so this has to be carefully investigated, and may not be possible.

### DIFF
--- a/src/PactNet.Abstractions/Verifier/Messaging/IMessageScenarioBuilder.cs
+++ b/src/PactNet.Abstractions/Verifier/Messaging/IMessageScenarioBuilder.cs
@@ -17,29 +17,29 @@ namespace PactNet.Verifier.Messaging
         IMessageScenarioBuilder WithMetadata(dynamic metadata);
 
         /// <summary>
-        /// Set the action of the scenario
+        /// Set the content factory of the scenario. The factory is invoked each time the scenario is required.
         /// </summary>
         /// <param name="factory">Content factory</param>
         void WithContent(Func<dynamic> factory);
 
         /// <summary>
-        /// Set the content of the scenario
+        /// Set the content factory of the scenario. The factory is invoked each time the scenario is required.
         /// </summary>
         /// <param name="factory">Content factory</param>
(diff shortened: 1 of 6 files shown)


adamrodger:
One problem with making the messaging provider async is that you end up with this signature on `IMessagenScenarios`:

```cs
    /// <summary>
    /// Add a message scenario
    /// </summary>
    /// <param name="description">Scenario description</param>
    /// <param name="factory">Message content factory</param>
    IMessageScenarios Add(string description, Func<dynamic> factory);

    /// <summary>
    /// Add a message scenario
    /// </summary>
    /// <param name="description">Scenario description</param>
    /// <param name="factory">Message content factory</param>
    IMessageScenarios Add(string description, Func<Task<dynamic>> factory);
```

When you invoke that then it causes an issue:

```cs
// sync invocation
Scenarios.Add("foo", () => new Foo());

// async invocations
Scenarios.Add("foo", () => Task.FromResult(new Foo())); // hmmmm
Scenarios.Add("foo", async () => await CreateFooAsync()); // also hmmmm
```

Because of the use of `dynamic` then technically all of those invocations actually match the sync version, and the API breaks. I haven't tested that out yet but I've had that problem before when the initial redesign happened for 4.x and we wanted to support async with dynamics.

Issue #492, opened 2024-02-18
RFC: Pact Plugins


adamrodger:
**Is your feature request related to a problem? Please describe.**
The Pact ecosystem now supports plugins, and PactNet should be part of this. See https://docs.pact.io/plugins/quick_start

**Describe the solution you'd like**
The implementation details are entirely TBD currently, but some thoughts:

- It is not desirable to add more and more dependencies to PactNet and make the library bigger and bigger. Instead the plugins should be separate projects, e.g. in a `PactNet.Extensions.*` namespace format.
- Plugins rely on FFI calls, which means that they need to load the FFI library and it has to be the same instance and version that the rest of the process has loaded so that it writes to the same Pact file in the same way etc.
    - It was originally envisaged that plugins could add extensions methods to add appropriate functionality (e.g. a `WithProtobufContent` to mirror the existing `WithJsonContent` calls) but the FFI aspect probably makes this pretty awkward
    - One solution may be to refactor the FFI interop parts out of `PactNet` and into a dedicated `PactNet.Interop` (or similar name) so that any plugin projects could depend on this and on `PactNet.Abstractions` to add functionality in a non-breaking way. This refactor would be a breaking change upfront though so this idea would need to be properly thought out.
- Plugins are installed to the OS, which creates friction in PactNet itself because you would install e.g. `PactNet.Extensions.Protobuf` but that doesn't actually install the plugin, just the API support in PactNet. This can be confusing for users.

**Breaking Changes**
Depending on the implementation details above (TBD at the time of writing) this may involve breaking changes.

**Potential Downsides/Caveats**
<TBD>

**Describe alternatives you've considered**
<TBD>

**Additional context**

- https://github.com/pactflow/pact-protobuf-plugin
- https://docs.rs/pact_ffi/latest/pact_ffi/plugins/index.html
- https://github.com/pact-foundation/pact-net/issues/489#issuecomment-1938708606

mefellows:
Thanks for raising this!

Some comments:

> Instead the plugins should be separate projects, e.g. in a PactNet.Extensions.* namespace format.

Is the implication here that we need a namespace for each plugin (e.g. one for `protobuf` and another for `CSV` etc.)? Or are you referring to the "plugin" work being contained in a separate namespace?

For context on my line of questioning, at the moment, the languages that have implemented plugins have started with what you might call a "raw" interface:

[...]

PR #548, opened 2025-09-04
Add PactNet.Extensions.Grpc package for grpc plugin support

max-lancaster:
Stacked on #551. Everything except the last commit belongs to #551 and will drop out of this diff when that PR merges. The remaining commit adds the `PactNet.Extensions.Grpc` package, which is the reference example of a wrapper library built on the raw plugin API described in docs/plugin-interactions.md.

I have read through issue #492 thoroughly and examined the source code and feel that I have a decent understanding of what's involved in adding plugin support, specifically for Grpc.  I was able to take @YOU54F's example and create this PR to get some sample Grpc tests working.

I have extracted a `PactNet.Interop` package as @adamrodger suggested in Issue #492 . This was not a breaking change since all the classes in this namespace were internal, `PactNet` now references this new package and none of the pre-existing interface methods have changed.  Extracting this package made it possible to create bare-bones Grpc plugin tests and is effectively a very "raw" way of using any plugin.  Perhaps a next step would be to wrap the bare-bones methods in a more fluent "raw" interface in a plugin package similar to implementations in other languages as @mefellows suggested.

Based on above "raw" tests I was able to refactor and extract shared functionality to create a new `PactNet.Extensions.Grpc` library that provides a fluent interface similar to the interfaces for http / rest and asynchronous messaging interactions.

An alternative approach here would be to just add `<InternalsVisibleTo Include="PactNet.Extensions.Grpc" />` to the PactNet.csproj file but this allows integration of other plugins in repos outside this one so it depends on whether you want to explicitly allow that or not.


adamrodger:
Many thanks for the PR - I can see you've put a lot of work into this!

I think a good path forward would be to create some kind of framework for plugins to be able to hook into existing PactNet internals without making those internals themselves public. A few examples of where the approach in this PR isn't desireable:

- The `PactNet.Interop` package has to have an `InternalsVisibleTo("PactNet.Extensions.Grpc")`. That means that developing a plugin wrapper effectively means that it has to be mainlined, whereas I'd like the option to allow other people to add plugin wrappers without mainlining
- The `GrpcPactBuilder` has to duplicate a lot of the functionality that already exists internally, for example the pattern around ensuring the mock server is started and then displaying the log messages afterwards. Plugin wrapper authors shouldn't need to do this, and PactNet itself should be the single place where things like log formatting are done so that the entire ecosystem has a consistent presentation.
- New classes have to be created, such as `PactFileWriter`, and these are either only accessible to mainlined plugin wrappers (undesireable as explained above) or need to be made public. If they're public then that creates a whole extra burden for us in ensuring they're totally safe to use at all times without corruping the FFI state, given we lose control of when they're being called.
- You can update the Interop package without updating the others - this means we can't verify compatibility as and when FFI features change.

For those reasons and more, I'd prefer to see some kind of framework created, probably as part of `PactNet` or maybe `PactNet.Abstractions` whereby plugin authors can register their plugins and add interactions to pacts (like the `UsePlugin` and `NewSyncMessageInteraction` methods you've added, which are a good addition) but cope with other internal details that plugin wrapper authors shouldn't need to worry about - writing the pact, starting the mock server, checking the response codes, displaying mismatches, freeing resources, etc.

To me, success would look like:

- It's possible to have a totally separate repository which can add a plugin wrapper, without any changes to PactNet itself. This repository would also contain the documentation and samples for how to get it working.
- It's possible to use a more "raw" plugin system in case a wrapper project doesn't exist, so that emerging new plugins don't need to wait until a nice wrapper is mainlined.
- We'd have some kind of documentation/guidance/workshop to show how a plugin wrapper can be written.

I think this is a great step to achieving that, and congratulations to you for this being the first time that we've been able to run plugins in PactNet. It's been a long time coming 😁

max-lancaster:
> * It's possible to have a totally separate repository which can add a plugin wrapper, without any changes to PactNet itself. This repository would also contain the documentation and samples for how to get it working.

@adamrodger I rebased this PR on top of PR #551 and made appropriate changes here to demonstrate that this is possible now. With those changes we can now create the new `PactNet.Extensions.Grpc` library in this repo or another repo as you prefer.

The only changes in this PR that are not in PR #551 are in https://github.com/pact-foundation/pact-net/pull/548/commits/aba5d3d713d033513301610c389bf148e28e74f7 which only adds the new library and associated tests.

> * It's possible to use a more "raw" plugin system in case a wrapper project doesn't exist, so that emerging new plugins don't need to wait until a nice wrapper is mainlined.

This was achieved in PR #551 for synchronous plugin interactions, I was thinking we could add asynchronous interactions in a separate PR.

> * We'd have some kind of documentation/guidance/workshop to show how a plugin wrapper can be written.

I'll leave it up to you how you want to go about doing this but would be willing to chip in if you want to provide some guidance.

timo-reymann:
@max-lancaster thanks for the effort already put into this. It would really unblock my work integrating Pact with our internal gRPC services. Happy to provide early feedback on usage etc. if needed - feel free to ping me.

jorgalad:
@max-lancaster thank you for working on this, from my side also happy to provide early testing

mefellows:
@adamrodger what's required to move forward with this? It looks like a great and substantial change for the ecosystem!

adamrodger:
@max-lancaster @mefellows the first stage is getting the other mentioned PR merged so that we can create a generic plugin interface, and then adding a gRPC plugin would be the first to adopt it (and prove that we'd made it appropriately generic hopefully also).

I'm having a look at that other PR now though.

jorgalad:
Hi @mefellows and @adamrodger, what's the current startus of this? We're currently trying to plan our involvement and how we'd like to continue with gRPC contract tests, would love to get an update if possible

mefellows:
It looks like #551 is in decent shape now, which I think is the crucial change required to support gRPC (and other plugins). Let's see what Adam comes back with.

PR #551, opened 2025-10-07
Add Synchronous Plugin Interaction Support 

max-lancaster:
@adamrodger I took your commit in the dev/plugins branch that you referenced in Issue #492  and made a "raw" plugin implementation that should work with the `PactNet.Extensions.Grpc` package I created in PR #548.

There may still be some design issues to consider.  One thing to note is that Plugins can have both synchronous and asynchronous interactions so I made my builder specific to the synchronous use case so we can add an asynchronous version at some point.


adamrodger:
I've added a commit for some cleanups and things just to save on a bunch of review comments.

I think the big design change that stands out is the two interfaces that have moved from the PactNet project to PactNet.Abstractions and made public instead of internal, and then making the port public also. Those are supposed to be implementation detail concerns so I'd prefer to keep them where they are.

We should provide some kind of generic public API if needed for plugins if they need to start a mock server instead of PactNet starting it for them at the appropriate time (but I don't see at first hand why that should happen vs. PactNet taking care of it).

max-lancaster:
> I think the big design change that stands out is the two interfaces that have moved from the PactNet project to PactNet.Abstractions and made public instead of internal, and then making the port public also. Those are supposed to be implementation detail concerns so I'd prefer to keep them where they are.

These changes were introduced in order to support a public `AbstractPactBuilder` to [share the pact verification functionality with extension library implementations](https://github.com/max-lancaster/pact-net/pull/1/changes#diff-c3045cef03aa866b2589c00c94b528c5a95776a28cb8a236586a6ca48a4d8716R21).  The verification functionality requires this and therefore these interfaces need to be made public in order to allow consuming libraries to use the verification functionality.

max-lancaster:
> I think the big design change that stands out is the two interfaces that have moved from the PactNet project to PactNet.Abstractions and made public instead of internal, and then making the port public also. Those are supposed to be implementation detail concerns so I'd prefer to keep them where they are.

Ok, @adamrodger  I took another look at this and I was able to revert the changes that made these details public and still keep my Grpc Extensions library working as intended, I think this should be ready to go now.  Basically I didn't really need to inherit from `AbstractPactBuilder` to make the `PactNet.Extensions.Grpc` library work.

mefellows:
Thanks for the great work here @max-lancaster - what's needed to move this one forward @adamrodger ?

src/PactNet.Abstractions/ISynchronousPluginRequestBuilder.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="name">Parameter name</param>
+        /// <param name="value">Parameter value</param>
+        ISynchronousPluginRequestBuilderV4 Given(string description, string name, string value);
+
+        /// <summary>
+        /// Add plugin interaction content
+        /// </summary>
+        /// <param name="contentType">Content type</param>
+        /// <param name="content">A dictionary containing the plugin content.</param>
+        ISynchronousPluginRequestBuilderV4 WithContent(string contentType, Dictionary<string, object> content);
```
adamrodger, line 30: question: Given this returns the same object again, what happens if this is called twice? Currently this allows:

```csharp
pact.UponReceiving("a request").WithContent("foo", foo).WithContent("bar", bar);
```

Is that intentional? Are we trying to allow setting different content types with different contents? What if we use the same content type but a different body? I think it's ambiguous whether those would combine or replace.

Other builders 'consume' the request builder when this part is called, so that you can't accidentally call it twice and thuse you avoid that ambiguity.

So I suggest maybe this returns some other type which then moves you onto the verification stage, then this has to be the last call and it can only happen once.

adamrodger (changes requested):
Just one API design choice to review, and then I think this is good to merge

src/PactNet.Abstractions/ISynchronousPluginRequestBuilder.cs
mefellows: On the request, this will replace the previous contents. 

It is allowed on the response, and will append the contents (i.e. multiple responses).

So on this method, it should "consume" the builder as you describe.

max-lancaster: Thanks @adamrodger & @mefellows, this is now addressed.

`WithContent` returns `void` and is the last call on the request builder. It can only be called once per interaction, so the ambiguity @adamrodger raised is gone. `Given` is still chainable before it.

I chose `void` rather than a new response stage after reading how `pactffi_interaction_contents` treats synchronous message interactions. For those, the FFI ignores the `InteractionPart` argument. The plugin receives one contents document and splits it into the request and the responses itself. Calling it twice replaces the request and appends a second response, exactly as @mefellows described. That means there is no separately addressable "response side" to build a stage around, and verification happens on the pact builder rather than per interaction, so the chain just ends.

Also in this update:

* unit tests for the plugin request builder and pact builder
* docs/plugin-interactions.md, covering the raw plugin API and how a wrapper library layers on top of it (the guidance Adam asked for on #548)
* rebased onto current master, and the PR base changed from dev/plugins to master so this can merge on its own

#548 remains stacked on this branch and will shrink to a single commit once this merges.

mefellows: Thanks Max! Over to you Adam

PR #559, opened 2026-08-13
feat: Add AsyncAPI reference support for message pacts (#557)

### LINKED ISSUE #557: [Feature Request] Add support for AsyncAPI references in Message Pact (.reference() method)
## Problem

When using PactNet to create Message Pact contracts for AsyncAPI Bi-Directional Contract Testing (BDCT) with Pactflow, there's currently no built-in way to link Message Pact interactions to AsyncAPI operations via the `comments.references.AsyncAPI` structure.

According to [PactFlow's AsyncAPI BDCT documentation](https://github.com/pactflow/example-bi-directional-provider-asyncapi), consumer pact files should include operation references like this:

  ```json
  {
    "description": "Message interaction",
    "comments": {
      "references": {
        "AsyncAPI": {
          "operationId": "receiveUserEvents"
        }
      }
    }
  }
```

## Current State in PactNet

PactNet 5.0.1 provides `IMessageBuilderV4.WithMetadata()` but does not expose a `.reference()` method to add AsyncAPI operation references.

## Workaround (Verified Working)

I manually added the `comments.references.AsyncAPI` structure to the generated pact file:

```json
  {
    "interactions": [
      {
        "description": "Some interaction",
        "comments": {
          "references": {
            "AsyncAPI": {
              "operationId": "someOperation"
            }
          }
        },
        "metadata": { ... },
        "contents": { ... }
      }
    ]
  }
```
After adding this structure manually, PactFlow successfully validated the Message Pact against the AsyncAPI provider specification. This confirms that the feature works end-to-end — PactNet just needs a programmatic way to add these references.

## Expected Behavior

PactNet should support something like this:

```csharp
  _pactBuilder
      .ExpectsToReceive("Some message")
      .Reference("AsyncAPI", "operationId", "someOperation")  // ← Missing
      .WithJsonContent(new { ... })
      .Verify<Message>(...)
```
  Or alternatively:

```csharp
  _pactBuilder
      .ExpectsToReceive("Some message")
      .WithAsyncApiReference("someOperation")  // ← Missing
      .WithJsonContent(new { ... })
      .Verify<Message>(...)
```
## Use Case

This feature is essential for teams using PactFlow's AsyncAPI BDCT approach with .NET Message Pact consumers. Without it, the generated pact files cannot be automatically validated against AsyncAPI provider specifications in Pactflow.

## Environment
- PactNet Version: 5.0.1
- .NET Version: 10.0
- Target: PactFlow AsyncAPI BDCT

Thanks in advance!

mr-roman-vivint:
Closes #557

## Summary
Implements AsyncAPI reference support for message pacts (#557)

## Problem
PactNet lacked a way to programmatically add AsyncAPI operation references to message interactions, which is required for PactFlow BDCT with AsyncAPI.

## Solution
Added three new methods to IMessageBuilderV4:
- `WithComment(key, value)` - for custom key-value comments
- `WithTextComment(comment)` - for text comments array
- `WithAsyncApiReference(operationId)` - convenience method for AsyncAPI refs

## Changes
- Core implementation in MessageBuilder and drivers
- FFI bindings for pactffi_set_comment and pactffi_add_text_comment
- Unit, integration, and E2E tests
- Updated sample code and documentation
- V4 only (V3 doesn't support comments per spec)

## Testing
All 411 tests passing. Added comprehensive test coverage.

## Breaking Changes
None - additive change to IMessageBuilderV4 only.


src/PactNet.Abstractions/IMessageBuilder.cs
```diff
@@ -75,6 +75,32 @@ public interface IMessageBuilderV4
...
+        /// <summary>
+        /// Add an AsyncAPI operation reference to the interaction
+        /// </summary>
+        /// <param name="operationId">the AsyncAPI operation ID</param>
+        /// <returns>Fluent builder</returns>
+        /// <remarks>
+        /// If called multiple times, only the last operationId will be retained.
+        /// For multiple reference types, use WithComment() with a custom JSON structure.
+        /// </remarks>
+        IMessageBuilderV4 WithAsyncApiReference(string operationId);
```
mefellows: I'd suggest changing this to something like

```suggestion
        IMessageBuilderV4 WithReference(string group, string name, string value);
```

src/PactNet/MessageBuilder.cs
```diff
@@ -119,6 +131,49 @@ internal MessageBuilder WithMetadata(string key, string value)
...
+        /// <returns>Fluent builder</returns>
+        /// <remarks>
+        /// If called multiple times, only the last operationId will be retained.
+        /// For multiple reference types, use WithComment() with a custom JSON structure.
+        /// </remarks>
+        internal MessageBuilder WithAsyncApiReference(string operationId)
+        {
+            // Create the nested JSON structure: { "AsyncAPI": { "operationId": "..." } }
+            string referencesJson = $"{{\"AsyncAPI\":{{\"operationId\":\"{operationId}\"}}}}";
+            this.driver.WithComment("references", referencesJson);
```
mefellows, line 149: Technically, this would work actually and would allow you to do what's needed without bumping to a newer FFI version, although if there are other comments this might replace them (I think).

It is also supported by the `pactffi_add_interaction_reference` function which surgically adds that reference into any existing comments.

See https://github.com/pact-foundation/pact-js-core/pull/861/changes for addition of the FFI functions to enable it.

mefellows (changes requested):
Some minor changes, with the primary one being the name of the reference.

I think we should make this a little less specific to AsyncAPI support. The underlying capability is really about adding arbitrary references to an interaction.

See example DSL for JS here: https://github.com/pact-foundation/pact-js/pull/1762/changes

It (references) could also be added to the HTTP interface, although that could be done in a separate PR if needed.

mefellows:
@adamrodger I've done a first pass at this based on the additions to other languages. I don't have any views on the .NET code themselves, so I'll leave that to you.

mr-roman-vivint:
@mefellows  Thank you for the feedback and sorry for the delay, a lot of work.

I've updated the implementation:
- Changed to a generic `WithReference(group, name, value)` method
- Removed `WithComment()` and `WithTextComment()` from the public API
- Added Dictionary-based state management to support multiple references per interaction
- Updated all tests.

Regarding `pactffi_add_interaction_reference`: I decided not to use it as it would require bumping the FFI version. I'm not familiar enough with the project to be confident about potential breaking changes that might not be covered by tests, so I kept the current FFI 0.4.27 and implemented it using `pactffi_set_comment` with state accumulation instead.

All tests pass. Let me know if you'd like any other changes.

Thanks in advance!

mefellows:
Thanks @mr-roman-vivint, this looks directionally better - I understand the reticence with bumping the FFI version. This is something that should be replaceable under the hood so the API doesn't change, but underlying implementation does.

Will leave the rest to @adamrodger (maintainer).

